PR #10, opened 2025-07-28
Fix enqueue_active_tasklet task lock guard not work

quininer:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=99a4fabd5706646c34a3c933eb48a27e

The behavior of `let _` is different from `let _guard`, which releases the variable immediately. This causes the lock to fail to protect against any operation.

I'm not really sure what this lock does, since the queue appears to be thread-safe. maybe this lock access can be remove.

### DIFF
--- a/kernel/src/asynk/mod.rs
+++ b/kernel/src/asynk/mod.rs
@@ -120,7 +120,7 @@ pub fn enqueue_active_tasklet(t: Arc<Tasklet>) {
         scheduler::current_thread_id()
     );
     let mut q = ASYNC_WORK_QUEUE.get_active_queue();
-    let _ = t.lock();
+    let _guard = t.lock();
     q.push_back(t.clone());
     #[cfg(debugging_scheduler)]
     crate::trace!(


Han-Jiang277 (approved):

lawkai-vivo:
Thanks for pointing it out. The lock was designed to be dropped at the end of this function.

> I'm not really sure what this lock does, since the queue appears to be thread-safe. maybe this lock access can be remove.

The queue is being protected by a lock. We still need another lock to protect the node inside the `Tasklet`, since this node might be accessed concurrently. Please check https://github.com/vivoblueos/kernel/blob/4adc3ebacdae0a274e85298698290b1a2505ca75/infra/src/tinyarc.rs#L160.

lawkai-vivo (approved):
LGTM.

lawkai-vivo:
I have grep a bit, looks there are several places using `let _ = lock()` idiom. @quininer Would you mind fixing them together?
```
logger.rs
66:        let _ = LOGGER_MUTEX.irqsave_lock();

boards/qemu_riscv64/uart.rs
139:        let _ = UART0_MUTEX.irqsave_lock();
156:        let _ = UART0_MUTEX.irqsave_lock();
169:        let _ = UART0_MUTEX.irqsave_lock();

asynk/mod.rs
123:    let _ = t.lock();
```

quininer:
> We still need another lock to protect the node inside the Tasklet, since this node might be accessed concurrently.

Makes sense, thank you.

> Would you mind fixing them together?

No, I'd rather just handle this one.

lawkai-vivo:
Opened #11 to fix remaining cases.

PR #14, opened 2025-08-05
fix: Recycle rx buffer after consuming RxToken in Virtio-Net device

ChenYunzhou:
Ensure the receive buffer is returned to the virtqueue after the RxToken is consumed, preventing virtqueue exhaustion and packet loss caused by unrecycled buffers.


lawkai-vivo (approved):
Since we haven't setup CI on github yet, I assume you have build and test locally.

PR #15, opened 2025-08-05
fix: Distinguish EOF from blocking mode in tcp socket

ChenYunzhou:
Introduce SocketError::WouldBlock for blocking state, keeping 0 exclusively for EOF to avoid ambiguity.


nooopslol (approved):
Since we haven't setup CI on github yet, I assume you have build and test locally.

PR #19, opened 2025-08-08
add start build ci

molovexue:
1.  Add "start build" CI functionality. Listen for comments in issues or PRs that contain "start build" and trigger the CI process (call the workflow from the public repository).

2. Create a public repository named org-common-workflows. Place the vivoblueos/org-common-workflows/.github/workflows/blueos-ci-reusable.yml@main file there to execute the CI.

3. Create a token using a bot account with repo and write:discussion permissions. Establish this as an Actions secret and variable for the organization. Name it REPO_ORCHESTRATOR_TOKEN01 with the aforementioned token value.

Test Repository: https://github.com/mogxtest/apps_shell/blob/main/.github/workflows/blueos-ci.yaml

### DIFF
--- a/.github/workflows/start_build-ci.yaml
+++ b/.github/workflows/start_build-ci.yaml
@@ -15,7 +15,7 @@ jobs:
         (github.event.issue.state == 'open') ||
         (github.event.pull_request.state == 'open')
       )
-    uses: vivoblueos/org-common-workflows/.github/workflows/blueos-ci-reusable.yml@main
+    uses: vivoblueos/toolchain/.github/workflows/blueos-ci-reusable.yml@main
     with:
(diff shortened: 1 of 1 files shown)


lawkai-vivo:
https://github.com/github/lock Could you please investigate https://github.com/github/lock? A global lock for merging PRs atomically is acceptable at the moment.

molovexue:
https://github.com/github/lock   This repository is primarily designed to lock the deployment environment and does not facilitate cross-repository merge operations. For merging across multiple repositories, I would recommend using the GitHub API to implement a best-effort mechanism. This includes rollback in case of merge failures. However, it cannot guarantee that developers will never encounter inconsistent repository states during the merging process

PR #20, opened 2025-08-13
fix: InterruptGuard lives not long enough to protect concurrent access

### LINKED ISSUE #11: LockGuard lives not long enough to protect concurrent access
This is follow-up of #10 . More details can be found on https://vojtechkral.github.io/blag/rust-drop-order/.

rust-cooker:
Description: InterruptGuard lives not long enough to protect concurrent access

Reason: use _ the guard returned by lock() is dropped right away on that first line and the lock isn't locked when resource is accessed.

fixes #11

### DIFF
--- a/kernel/src/scheduler/mod.rs
+++ b/kernel/src/scheduler/mod.rs
@@ -421,7 +421,7 @@ pub(crate) extern "C" fn schedule() -> ! {
 
 #[inline]
 pub fn current_thread() -> ThreadNode {
-    let _ = DisableInterruptGuard::new();
+    let _dig = DisableInterruptGuard::new();
     let my_id = arch::current_cpu_id();
     let t = unsafe { RUNNING_THREADS[my_id].assume_init_ref().clone() };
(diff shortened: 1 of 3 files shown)


lawkai-vivo:
This isssue is fixed by #18

lawkai-vivo:
Ok, I see #18 is partial fix.

lawkai-vivo:
Please rebase this PR.

kernel/src/scheduler/mod.rs
```diff
@@ -421,15 +421,15 @@ pub(crate) extern "C" fn schedule() -> ! {
 
 #[inline]
 pub fn current_thread() -> ThreadNode {
-    let _ = DisableInterruptGuard::new();
+    let _dig = DisableInterruptGuard::new();
```
han-jiang277: _guard is better

zwbZ: Good suggestion

PR #24, opened 2025-08-14
Fix: Prevent broken in MPS2 coverage collection

AbeZbm:


lawkai-vivo:
Does it apply to qemu_mps3_an547?

AbeZbm:
I have tested the same scenario on qemu_mps3_an547 and confirmed that the issue does not appear on MPS3. Therefore, this change is only required for MPS2.

lawkai-vivo:
@AbeZbm Would you mind signing the CLA?

AbeZbm:
Sorry, I had signed the CLA before, but my previous commit used the wrong user.name and user.email, so it wasn’t linked to my account. I’ve fixed the author info and force-pushed.

PR #28, opened 2025-08-19
fix: Use SpinLock to protect the GlobalQueue

lawkai-vivo:


nooopslol (approved):
In case GLOBALQUEUE will be used in the irq handler later

PR #30, opened 2025-08-19
feat: Introduce UniqueListHead and AtomicListHead to enhance concurrent control

lawkai-vivo:
By exploiting Rust's zero overhead safe abstraction, introduce `UniqueListHead` to model ListHead belonging to only one static list.

By exploiting tagged pointer, introduce `AtomicListHead` to implement concurrent safe list operations.


lawkai-vivo:
Rebased, will wait for the CI.

PR #35, opened 2025-08-29
Add support for event flags

Han-Jiang277:


kconfig/config/qemu_mps2_an385/coverage/defconfig
```diff
@@ -1,41 +1,35 @@
+CONFIG_IRQ_PRIORITY_BITS_2=y
+# CONFIG_IRQ_PRIORITY_BITS_3 is not set
+# CONFIG_IRQ_PRIORITY_BITS_8 is not set
 CONFIG_ALIGN_SIZE=8
-CONFIG_TICK_PER_SECOND=100
-CONFIG_MAIN_THREAD_PRIORITY=100
+CONFIG_TICKS_PER_SECOND=100
```
lawkai-vivo, line 5: Please post another PR for this irrelevant change.

kernel/src/thread/mod.rs
```diff
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 extern crate alloc;
+#[cfg(event_flags)]
```
lawkai-vivo, line 16: Why do we need this flag? I'm imaging this should apply to all archs we are supporting.

kernel/src/sync/event_flags.rs
```diff
@@ -0,0 +1,514 @@
...
+    pub struct EventFlagsMode: u8 {
+        const ANY = 1 << 0;
+        const ALL = 1 << 1;
+        const NO_CLEAR = 1 << 2;
+    }
+}
+
+#[derive(Debug)]
+pub struct EventFlags {
+    // We let the Spinlock protect the whole semaphore.
```
lawkai-vivo:
```suggestion
    // We let the Spinlock protect the whole EventFlags.
```

kernel/src/sync/event_flags.rs
```diff
@@ -0,0 +1,514 @@
...
+}
+
+impl Default for EventFlags {
+    fn default() -> Self {
+        Self::const_new()
+    }
+}
+
+impl EventFlags {
+    pub const fn const_new() -> Self {
```
lawkai-vivo:
```suggestion
    pub const fn new() -> Self {
```
We use `const_new` when we have two versions of constructor, `const_new` for static construction, `new` for runtime construction. If `new` is already `const`, there is no need to add `const_new`. We prefer using `new` first since it's common in Rust code.

kernel/src/sync/event_flags.rs
```diff
@@ -0,0 +1,514 @@
...
+        // Set some flags
+        assert!(event_flags.set(0x01).is_ok());
+        assert!(event_flags.set(0x02).is_ok());
+        assert_eq!(event_flags.get(), 0x03);
+
+        // Clear flags that partially exist
+        event_flags.clear(0x05); // Clear 0x01 and 0x04 (0x04 doesn't exist)
+        assert_eq!(event_flags.get(), 0x02); // Only 0x02 remains
+    }
+}
```
lawkai-vivo, line 517: Please also add tests for SMP.

kernel/src/thread/mod.rs
han-jiang277: it offers the advantage of reducing memory usage in scenarios where event_flags are not required

kconfig/config/qemu_mps2_an385/coverage/defconfig
han-jiang277: it's not a irrelevant change. we need to rerun menuconfig after add a new config, and copy .config to debug, converage and release defconfig

kernel/src/sync/event_flags.rs
han-jiang277: add a  stress test for event_flags in kernel lib.rs

nooopslol (approved):

PR #38, opened 2025-09-02
Increase infra crate coverage by adding and updating tests

AbeZbm:
Hi,

Thank you for reviewing this PR.

We are researchers focusing on generating Rust tests using LLMs to satisfy specific execution paths in functions. While examining the existing code, we identified opportunities to add and refine tests to improve the `infra` crate’s overall coverage. We have added new unit tests and updated several existing ones, and all tests have been carefully reviewed to ensure correctness and effectiveness.

- `infra` unit tests line coverage before: infra/src: [80.25%], infra/src/list: [77.25%]
- `infra` unit tests line coverage after:  infra/src: [87.32%], infra/src/list: [90.02%]

### DIFF
--- a/infra/src/list.rs
+++ b/infra/src/list.rs
@@ -15,3 +15,7 @@
 pub mod singly_linked_list;
 pub mod typed_atomic_ilist;
 pub mod typed_ilist;
+
+pub trait GenericList {
+    type Node;
+}
--- a/infra/BUILD.gn
+++ b/infra/BUILD.gn
@@ -23,6 +23,7 @@ group("check_infra") {
 shared_deps = [
   "//external/cfg-if/v1.0.0:cfg_if",
   "//external/memchr/v2.7.4:memchr",
+  "//external/spin/v0.9.8:spin",
 ]
 
 build_rust("blueos_infra") {
--- a/infra/src/list/singly_linked_list.rs
+++ b/infra/src/list/singly_linked_list.rs
@@ -378,6 +378,7 @@ mod tests {
             count += 1;
         }
         assert_eq!(count, 1023);
+        assert!(head.remove().is_none());
     }
 
     #[bench]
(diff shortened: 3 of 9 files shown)


infra/src/list.rs
```diff
@@ -15,3 +15,7 @@
 pub mod singly_linked_list;
 pub mod typed_atomic_ilist;
 pub mod typed_ilist;
+
+pub trait GenericList {
```
nooopslol, line 19: please add the benefit of GenericList, or add test cases to reflect it.

AbeZbm: Thanks for reviewing my pr.  

The benefit of GenericList trait is that it provides a unified abstraction for intrusive lists, allowing structures to attach to multiple lists without conflicts and enabling generic utilities/tests to work across different list implementations.

lawkai-vivo: Another point is Rust doesn't support
```
impl Foo {
  type Bar = XXX;
}
Foo::Bar // <- doesn't support.
```

AbeZbm: Thanks for the feedback.
In addition to the conceptual benefit, the current tests already reflect the role of GenericList. For example, many cases ([test_insert_and_detach](https://github.com/vivoblueos/kernel/blob/f7f965f0d4017960034575cab97e7d99d1652880/infra/src/tinyarc.rs#L691), [test_push_and_pop](https://github.com/vivoblueos/kernel/blob/f7f965f0d4017960034575cab97e7d99d1652880/infra/src/tinyarc.rs#L733), [insert_by_prio](https://github.com/vivoblueos/kernel/blob/f7f965f0d4017960034575cab97e7d99d1652880/infra/src/tinyarc.rs#L957), etc.) explicitly use ```<ControlStatusList as GenericList>::Node``` as ```type L``` and perform insert/detach/iterate operations on it.
These tests validate that different intrusive lists can be accessed in a uniform way through the GenericList abstraction, which is exactly the purpose of the trait.

https://github.com/vivoblueos/kernel/blob/f7f965f0d4017960034575cab97e7d99d1652880/infra/src/tinyarc.rs#L611-L616

nooopslol (approved):

PR #54, opened 2025-09-22
feat: Run only one test case

lawkai-vivo:
To use this feature, developer has to declare `test_only` at the beginning of the crate.
```
blueos_test_macro::test_only!();
```
Then change the interested test case from `#[test]` to `#[only_test]`. All `#[test]` will be disabled and only the first `#[only_test]` is enabled.

This can help us debug test case easier.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/54

PR #66, opened 2025-10-14
[CI] Fix string comparison

lawkai-vivo:


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/66

PR #77, opened 2025-10-14
[CI] Adjust step order

lawkai-vivo:
Put applying patches before freeing spaces, so that we can know if conflicts happen earlier.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/77

PR #83, opened 2025-10-15
Remove redundant targets

lawkai-vivo:
These targets are no longer used.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/83

han-jiang277 (approved):
LGTM

PR #72, opened 2025-10-14
Run shell in bcm2711

zhanghe-vivo:
Description: N/A

Reason: N/A

Issue: N/A

PR #88, opened 2025-10-15
Revert "Run shell in bcm2711"

lawkai-vivo:
Reverts vivoblueos/kernel#72 which disabled network subsystem and broke CI.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/88

PR #89, opened 2025-10-16
Add cmsis ticks functions

zhanghe-vivo:


han-jiang277:
build_prs https://github.com/vivoblueos/kernel/pull/89

PR #97, opened 2025-10-17
Add use_cmsis_adapter in gn

han-jiang277:


adapter/cmsis/BUILD.gn
```diff
@@ -16,65 +16,67 @@ import("//build/templates/build_template.gni")
...
-  edition = "2021"
-  proc_macro_deps = [
-    "//external/rust-delegate/v0.13.3:delegate",
-    "//kernel/test_harness:blueos_test_macro",
-  ]
-  deps = adapter_deps
-  deps += [
-    ":cmsis_adapter",
-    "//external/semihosting/v0.1.20:semihosting",
+if (use_cmsis_adapter) {
```
bzEq: Why do we need this guard? It's already guarded in adapter/BUILD.gn

han-jiang277: do not build cmsis when cmsis is not configured in Kconfig, like aarch64 and riscv64

bzEq: I mean it's already guarded by the parent BUILD.gn. No need to check it twice.

lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/97

adapter/cmsis/BUILD.gn
han-jiang277: That makes sense, let me try it.

han-jiang277:
Job run with https://github.com/vivoblueos/build/pull/13 is passed.

PR #98, opened 2025-10-17
[CI] Add Re-sync step as manifest may changed

han-jiang277:


han-jiang277:
build_prs https://github.com/vivoblueos/kernel/pull/98

PR #107, opened 2025-10-29
Add experimental support of signal

lawkai-vivo:
Entering signal context when the thread is re-scheduled and has pending signals. Thread uses the same stack in signal context and the signal handler is run in thread mode. Before entering the signal context, thread context is saved and is restored on exiting of the signal context.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/107

lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/107

PR #116, opened 2025-11-03
Support boot with MSP in arch/arm

han-jiang277:
Support boot with provided MSP address.
IsNotNullshould be replaced with is_null.

### DIFF
--- a/infra/src/lib.rs
+++ b/infra/src/lib.rs
@@ -37,36 +37,3 @@ pub mod spinarc;
 pub mod string;
 pub mod tinyarc;
 pub mod tinyrwlock;
-
-pub trait IsNotNull {
-    fn is_not_null(&self) -> bool;
-}
-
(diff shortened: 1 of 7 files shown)


han-jiang277:
build_prs https://github.com/vivoblueos/kernel/pull/116

han-jiang277:
build_prs https://github.com/vivoblueos/kernel/pull/116

infra/src/lib.rs
```diff
@@ -37,36 +37,3 @@ pub mod spinarc;
 pub mod string;
 pub mod tinyarc;
 pub mod tinyrwlock;
-
```
lawkai-vivo, line 40: Why do we need to revert this?

han-jiang277: Since Rust doesn't natively provide is_not_null, using !is_null directly would be more consistent.

PR #119, opened 2025-11-04
Split ffi module to a source file

han-jiang277:


han-jiang277:
build_prs https://github.com/vivoblueos/kernel/pull/119

PR #122, opened 2025-11-04
Fix set_origin_priority in thread build

han-jiang277:


han-jiang277:
build_prs https://github.com/vivoblueos/kernel/pull/122

lawkai-vivo (approved):

PR #124, opened 2025-11-05
fix: Spin until we actually get the lock of the node

lawkai-vivo:
We can't tell if locking fails or the action isn't actually performed. Perform spinning before taking action on the node.

We require users must acquire the list's lock first, often `spinlock.irqsave_lock()`, so we don't enforce disabling local irq before locking the list node here.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/124

lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/124

PR #125, opened 2025-11-05
Make resetting test threads more robust

lawkai-vivo:
Every test case uses different counters for waiting test threads end.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/125

PR #142, opened 2025-11-18
[RISCV] Adjust bootstrap code to init global pointer

lawkai-vivo:


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/142

PR #156, opened 2025-11-19
Skip manifests when checking license

lawkai-vivo:


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/156

PR #152, opened 2025-11-19
Adjust implementation of RelativeAdapter

lawkai-vivo:
Make it more clear and easier to reason.

PR #158, opened 2025-11-19
Fix regression introduced by #152

lawkai-vivo:
Fix https://github.com/vivoblueos/build/actions/runs/19501014930/job/55815063338.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/158

PR #183, opened 2025-11-25
Fix: ensure CI fails when format check fails

wangchen7722:
https://github.com/vivoblueos/kernel/actions/runs/19659006965/job/56301638524

gn format exits with code 2, while yapf3, rustfmt, and yamlfmt exit with code 1 and print diffs to stdout. 

Previously, failures from gn and rustfmt were skipped because their error messages did not contain the keyword `format. 
This patch fixes that behavior and keeps non-format command errors skipped as before.

### DIFF
--- a/.github/workflows/reusable_cross_repos_build.yml
+++ b/.github/workflows/reusable_cross_repos_build.yml
@@ -164,7 +164,6 @@ jobs:
                 console.log(`Skipping ${repo} for format check`);
                 continue;
               }
-              console.log(`\nProcessing: repo=${repo}, file=${file}`);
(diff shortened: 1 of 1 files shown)


wangchen7722:
build_prs https://github.com/vivoblueos/kernel/pull/183

wangchen7722:
build_prs https://github.com/vivoblueos/kernel/pull/183

wangchen7722:
build_prs https://github.com/vivoblueos/kernel/pull/183

han-jiang277 (approved):

PR #200, opened 2025-12-03
Adjust some atomic orderings to be less strict

lawkai-vivo:
Improve ~3% in thread_metric benchmark.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/200

PR #202, opened 2025-12-04
Remove posix mod which is unused

lawkai-vivo:


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/202

han-jiang277 (approved):

PR #203, opened 2025-12-05
Detach WaitEntry from the WaitQueue if timeout occurs

lawkai-vivo:
Also move insertion of the `WaitEntry` out of `suspend_me_with_timeout`.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/203

PR #212, opened 2025-12-10
Add LifetimeDelegator to track lifetimes of values "borrowed" via pointers

lawkai-vivo:
This is useful if we are appending stack objects to our `ListHead` which doesn't feature lifetime management like `ArcList`.
Since the `LifetimeDelegator` is ZST, there's no runtime overhead.

### DIFF
--- a/infra/src/lifetime.rs
+++ b/infra/src/lifetime.rs
@@ -15,7 +15,7 @@
 use core::marker::PhantomData;
 
 #[derive(Default)]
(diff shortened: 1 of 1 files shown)


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/212

lawkai-vivo:
prefer https://github.com/vivoblueos/kernel/pull/212

PR #224, opened 2025-12-17
Bugfix: memory may used by other core in SMP mode. 

han-jiang277:
delete test for memory_info.


han-jiang277:
build_prs https://github.com/vivoblueos/kernel/pull/224

PR #233, opened 2025-12-18
Introduce LiveFor, Iou and IouMut

lawkai-vivo:
These types are used to track lifetimes of values accessed via pointers.

The `LiveFor` doesn't incur runtime overhead.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/233

PR #236, opened 2025-12-19
Remove const_new API

lawkai-vivo:
If `new` and `const_new` are different, `const_new` is renamed.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/236

han-jiang277 (approved):

PR #249, opened 2025-12-25
Increase timeout of tm_preemptive_scheduling_test.checker

lawkai-vivo:
It sometimes timeout if the machine is at high workload.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/249

PR #259, opened 2025-12-30
Do not spin to lock local irq

lawkai-vivo:
This can save some cycles to get the lock.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/259

han-jiang277 (approved):

PR #265, opened 2026-01-04
Fix null pointer dereference

lawkai-vivo:
Fixes https://github.com/vivoblueos/kernel/issues/232.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/265

han-jiang277 (approved):

PR #267, opened 2026-01-04
Advance year in license

lawkai-vivo:
2025 is passed.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/267

PR #272, opened 2026-01-06
Remove fifo scheduler

lawkai-vivo:
The fifo scheduler is not under active maintenance.


lawkai-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/272

han-jiang277:
build_prs

PR #280, opened 2026-01-12
Modify FLASH size of gd32e507_eval in debug mode

xuchang-vivo:
In fact, the GD32E507's flash size cannot exceed 512 KB. Here it is set to 1024 KB — exceeding the actual hardware limit — to allow the firmware to compile so the emulator can run it for debugging. 

Note: the GD32E507 firmware built in debug mode will not run on real hardware


xuchang-vivo:
build_prs https://github.com/vivoblueos/build/pull/50 https://github.com/vivoblueos/kernel/pull/280

PR #285, opened 2026-01-14
Make statistics of thread optional

lawkai-vivo:
Stats of thread has large impact of the performance. Make it optional and disable it by default.
By disabling it, preemptive scheduling in thread_metric can be improved by ~40%.

### DIFF
--- a/adapter/cmsis/src/os1/tick.rs
+++ b/adapter/cmsis/src/os1/tick.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use blueos::time;
+use blueos::{scheduler, time};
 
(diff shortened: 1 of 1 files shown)


lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

adapter/cmsis/src/os1/tick.rs
```diff
@@ -35,7 +35,7 @@ mod tests {
 
     #[test]
     fn test_osKernelSysTick() {
+        // If the system runs fast enough, it can be zero.
```
han-jiang277, line 39: maybe add sleep(1) ?

lawkai-vivo:
build_prs

adapter/cmsis/src/os1/tick.rs
lawkai-vivo: Done

han-jiang277 (approved):

PR #288, opened 2026-01-15
Fix wrong assertion in global scheduler

lawkai-vivo:
We are keeping the invariance that iff a thread is at READY state, it's in the ready queue, except the idle thread.


lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

lawkai-vivo:
Fixed in #294

PR #290, opened 2026-01-15
Mask mcause for valid bits

xuchang-vivo:
The MCAUSE register only has bit 31 and bits [4:0] (or [3:0]) defined as valid in common; the other bits might be used in certain soc, and are not guaranteed to be zero.

See: https://docs.amd.com/r/en-US/ug1629-microblaze-v-user-guide/Machine-Cause-Register-mcause


xuchang-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/290

xuchang-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/290

lawkai-vivo:
What if we only initialize the `mcause` to 0 at boot stage?

xuchang-vivo:
> What if we only initialize the `mcause` to 0 at boot stage?

Some chips may assign additional meanings to other bits, so simply initializing them to 0 is inappropriate. Only bit31 and [0:5] are common in the spec.

lawkai-vivo (approved):

Issue #299, opened 2026-01-26
stress_build_threads occasionally fails on riscv32


lawkai-vivo:
```
[ RUN      ] stress_build_threads
panicked at ../../kernel/kernel/src/arch/riscv/trap.rs:223:9:
assertion failed: ok
Oops: assertion failed: ok
```
See https://github.com/vivoblueos/kernel/actions/runs/21345463315/job/61432111319
https://github.com/vivoblueos/kernel/blob/45adebd5e8ba08346e339c68ca26ab5c1dc91df2/kernel/src/arch/riscv/trap.rs#L223

PR #300, opened 2026-01-26
Fix incorrect state transfer when preemption is disabled

lawkai-vivo:
Consider following function call chain in rv32

0. yield_me_now_or_later
1. suspend_me
2. disable_preempt
3. timer interrupt occurs
4. claim pending switch and switch to another thread

So check is_preemptable in yield_me_now_or_later is invalid. We should check it when we are really deciding to switch or not.

Fixes https://github.com/vivoblueos/kernel/issues/299.


lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

PR #305, opened 2026-01-26
Add Clock trait and implementations

lawkai-vivo:


lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

kernel/src/devices/clock/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// We use the term `cycles` to refer to the internal counter of the Clock. We
+// use the term `hz` to descripe how many cycles within a second. Tick is
+// defined as a short period of time, which is atomic in the system, just like
+// the Planck time in physical world. We use `TICKS_PER_SECOND` to measure it.
+// A clock instance should be able to interrupt the system.
+pub trait Clock {
```
xuchang-vivo, line 20: Is there a missing interface to set the clock frequency?

lawkai-vivo: Clock frequency is fixed for kernel. If it's configurable and should be done by firmware not the kernel. The kernel expects a stable monotonic counter.

xuchang-vivo: There may be cases where the SoC frequency is adjusted based on the power/performance mode

lawkai-vivo: That's the CPU's frequency I guess. If we're entering low power mode, this clock should be stopped and use the low power timer.

lawkai-vivo:
build_prs

PR #316, opened 2026-01-29
Enable coop sched test

lawkai-vivo:


lawkai-vivo:
build_prs

lawkai-vivo:
We still need to disable RR to run the coop sched.

PR #319, opened 2026-02-02
Make signal handling after the thread is re-scheduled

lawkai-vivo:
We are making waitentry and timer on stack. To avoid dangling pointers when the thread is terminated forcely, we have to wake up the thread to detach from the waitqueue or sleepqueue. To waking up the thread, we'll use the signal mechanism.


lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

han-jiang277 (approved):

PR #321, opened 2026-02-02
Optimize performance of pop_front in ArcList

lawkai-vivo:
When we're popping the front value, we first cloned the arc and then decremented the arc. The two operations can be combined.


lawkai-vivo:
build_prs

lawkai-vivo:
build_prs

han-jiang277 (approved):

PR #336, opened 2026-02-25
Make MIE optional

xuchang-vivo:
MIE register is not always available in Riscv arch. So make it optional.


xuchang-vivo:
build_prs

PR #350, opened 2026-03-06
Provide a shared exception vector for cortex-m

xuchang-vivo:
On Cortex‑M series chips, there should be similar exception vectors.

### DIFF
--- a/kernel/src/arch/arm/mod.rs
+++ b/kernel/src/arch/arm/mod.rs
@@ -16,6 +16,8 @@ pub(crate) mod hardfault;
 pub mod irq;
 pub(crate) mod xpsr;
 use crate::{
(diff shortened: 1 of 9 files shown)


xuchang-vivo:
build_prs

lawkai-vivo:
What's the benefit of using a common exception handler? IMO, hardware dispatch should be faster than software dispatch.

xuchang-vivo:
> What's the benefit of using a common exception handler? IMO, hardware dispatch should be faster than software dispatch.

I set a misleading title — it should be “shared exception vector”, not “common exception handler”.

lawkai-vivo (approved):

han-jiang277 (approved):

PR #355, opened 2026-03-10
use config seperate librs/infra string functions

zhanghe-vivo:
add some missing string functions, but some of them  reuse same #[no_demagle] posix implementation , they conflict with infra implementations, now add an config object in GN as an solution

https://github.com/vivoblueos/librs/pull/28


infra/BUILD.gn
```diff
@@ -46,6 +46,13 @@ build_rust("blueos_infra") {
   deps = shared_deps
 }
 
+config("disable_string_exports") {
```
xuchang-vivo: Is it possible to set in Kconfig?

infra/src/string.rs
```diff
@@ -87,6 +87,7 @@ impl<T: Zero> NulTerminated<'_, T> {
 
 // used by core/src/slice/cmp.rs
 /// See <https://pubs.opengroup.org/onlinepubs/9799919799/functions/memcmp.html>.
+#[cfg(not(disable_infra_c_string_exports))]
```
han-jiang277: This function is already weak-linked. Why do we still need to isolate it based on cfg?

zhanghe-vivo: i can try to let  all of them  use weak-linked attributes, let linker choose one(should be order , maybe choose the first occurence）,  therefore, also don't need  change these build.gn files

PR #356, opened 2026-03-10
Fix print panic when run in MacOS

han-jiang277:


han-jiang277:
build_prs

han-jiang277:
build_prs

kernel/src/logger.rs
```diff
@@ -42,7 +42,7 @@ pub fn set_max_level(level: LogLevel) {
 pub fn logger_init() {
     static LOGGER: Logger = Logger {};
     #[cfg(debug)]
-    log::set_max_level(LevelFilter::Trace);
```
xuchang-vivo, line 45: Is it possible to set in kconfig

Issue #353, opened 2026-03-06
Test_mutex_multi_thread_timeout failed occasionally on RISCV


xuchang-vivo:
As follow:
```
[       OK ] test_mutex_multi_thread_priority1
[ RUN      ] test_mutex_multi_thread_timeout
panicked at ../../kernel/kernel/src/sync/mutex.rs:593:17:
assertion failed: !result
Oops: assertion failed: !result
```

xuchang-vivo:
https://github.com/vivoblueos/kernel/blob/a06ea55a9daa2eb2b2fc9f8fd41ebd0bd337d9ea/kernel/src/sync/mutex.rs#L592-L610
There is a possibility that the logic on line 592 is executed only after the logic on line 610 has finished executing.

PR #357, opened 2026-03-11
Add repeat attr to test harness

xuchang-vivo:
Usage:
```
    #[test(repeat=500)]
    fn test_mutex_multi_thread_timeout() {
```
Some unit tests may fail occasionally; using repeat can increase the likelihood of triggering the failure. 

See https://github.com/vivoblueos/kernel/issues/353


xuchang-vivo:
build_prs

PR #362, opened 2026-03-16
Aovid panic in calloc due to unaligned required size

xuchang-vivo:
https://github.com/vivoblueos/kernel/issues/310


xuchang-vivo:
build_prs

PR #370, opened 2026-03-17
fix(timer): relax timeout tolerance for QEMU virtualized environments

han-jiang277:
- Relax the timer timeout accuracy test tolerance from 10-12 to 10-20 ticks to accommodate timing variations in QEMU virtualized environments
- The test was failing because timer precision can be affected by host system load and CPU scheduling delays in virtualized environments

### DIFF
--- a/kernel/src/time/timer.rs
+++ b/kernel/src/time/timer.rs
@@ -551,7 +551,7 @@ mod tests {
         let now = Tick::now();
         let diff = now.0 - base.0;
         // Should be no diff for hard timer at low workload.
-        // Allow wider tolerance for QEMU virtualized environments where timer
+        // FIXME: Allow wider tolerance for QEMU virtualized environments where timer
         // precision can be affected by host system load and CPU scheduling.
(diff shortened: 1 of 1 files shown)


han-jiang277:
build_prs

han-jiang277:
build_prs

han-jiang277:
build_prs

kernel/src/time/timer.rs
```diff
@@ -551,7 +551,9 @@ mod tests {
         let now = Tick::now();
         let diff = now.0 - base.0;
         // Should be no diff for hard timer at low workload.
-        assert_eq!(diff, 10);
+        // Allow wider tolerance for QEMU virtualized environments where timer
```
xuchang-vivo: Adding a FIXME here might be better.

han-jiang277: Running QEMU with the -icount parameter would consume too much CI time, so there are no plans to do so for now.

han-jiang277:
build_prs

PR #375, opened 2026-03-22
fix: mutex timeout test and Tick::after timing calculation

han-jiang277:
## Summary
- Fix `Tick::after()` to call `Self::now()` only once instead of twice, avoiding potential timing issues
- Fix https://github.com/vivoblueos/kernel/issues/353
- Add QEMU virtualization timing tolerance in test

## Changes

1. **kernel/src/time.rs**: Fix `Tick::after()` - use saved `now` variable instead of calling `Self::now()` twice

2. **kernel/src/sync/mutex.rs**: 
   - Add detailed test documentation explaining the test logic
   - Add wait_before_release comment for QEMU timing tolerance

3. **kernel/src/time/timer.rs**: Minor cleanup (unused variable)

## Test plan
- [x] Run `test_mutex_multi_thread_timeout` - passes consistently
- [x] Run full CI test suite

### DIFF
--- a/kernel/src/sync/mutex.rs
+++ b/kernel/src/sync/mutex.rs
@@ -588,7 +588,7 @@ mod tests {
     /// Test logic:
     /// - Main thread acquires mutex and holds for ~10 ticks
     /// - Child thread tries to acquire mutex with 5 tick timeout (LESS than main holds)
-    /// - Child thread should timeout (result = true)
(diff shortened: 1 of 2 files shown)


han-jiang277:
build_prs

han-jiang277:
build_prs

PR #376, opened 2026-03-22
fix(time): avoid multiplication overflow in from_clock_cycles

han-jiang277:
Fixes the multiplication overflow issue in the from_clock_cycles function in kernel/src/time.rs.
```
panicked at ../../kernel/kernel/src/time.rs:114:15:
attempt to multiply with overflow
Oops: attempt to multiply with overflow
```
This patch ensures safe arithmetic when converting clock cycles to time values.

### DIFF
--- a/kernel/src/time.rs
+++ b/kernel/src/time.rs
@@ -66,8 +66,8 @@ impl Tick {
         debug_assert_eq!(ClockImpl::hz() % TICKS_PER_SECOND as u64, 0);
         // Use u128 to avoid multiplication overflow when cycles is large
         Self(
-            (ClockImpl::estimate_current_cycles() as u128 * TICKS_PER_SECOND as u128 / ClockImpl::hz() as u128)
-                as usize,
+            (ClockImpl::estimate_current_cycles() as u128 * TICKS_PER_SECOND as u128
+                / ClockImpl::hz() as u128) as usize,
         )
     }
 
@@ -83,7 +83,9 @@ impl Tick {
             return;
         }
         // Use u128 to avoid multiplication overflow when n is large
-        ClockImpl::interrupt_at((ClockImpl::hz() as u128 * n.0 as u128 / TICKS_PER_SECOND as u128) as u64);
+        ClockImpl::interrupt_at(
+            (ClockImpl::hz() as u128 * n.0 as u128 / TICKS_PER_SECOND as u128) as u64,
+        );
     }
 }
 
--- a/kernel/src/time.rs
+++ b/kernel/src/time.rs
@@ -64,11 +64,8 @@ impl Tick {
 
     pub fn now() -> Self {
         debug_assert_eq!(ClockImpl::hz() % TICKS_PER_SECOND as u64, 0);
-        // Use u128 to avoid multiplication overflow when cycles is large
-        Self(
-            (ClockImpl::estimate_current_cycles() as u128 * TICKS_PER_SECOND as u128
-                / ClockImpl::hz() as u128) as usize,
-        )
+        let cycles_per_tick = ClockImpl::hz() / TICKS_PER_SECOND as u64;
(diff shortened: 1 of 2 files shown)


han-jiang277:
build_prs

han-jiang277:
build_prs

han-jiang277:
build_prs

han-jiang277:
build_prs

kernel/src/time.rs
```diff
@@ -64,9 +64,10 @@ impl Tick {
 
     pub fn now() -> Self {
         debug_assert_eq!(ClockImpl::hz() % TICKS_PER_SECOND as u64, 0);
+        // Use u128 to avoid multiplication overflow when cycles is large
         Self(
-            (ClockImpl::estimate_current_cycles() * TICKS_PER_SECOND as u64 / ClockImpl::hz())
-                as usize,
+            (ClockImpl::estimate_current_cycles() as u128 * TICKS_PER_SECOND as u128
```
xuchang-vivo: The u64 type is sufficient here, as TICKS_PER_SECOND will not exceed 1000 and is usually 100. The u128 type would require four load operations on a 32-bit system, so it's not recommended for such a frequently used function as now.

kernel/src/time.rs
```diff
@@ -81,7 +82,10 @@ impl Tick {
             ClockImpl::stop();
             return;
         }
-        ClockImpl::interrupt_at(ClockImpl::hz() * n.0 as u64 / TICKS_PER_SECOND as u64);
+        // Use u128 to avoid multiplication overflow when n is large
+        ClockImpl::interrupt_at(
+            (ClockImpl::hz() as u128 * n.0 as u128 / TICKS_PER_SECOND as u128) as u64,
```
xuchang-vivo: This cannot overflow either; even if we assume hz equals the maximum value of u32 (which is 4.29GHz), then tick would need to equal 2**32-1 to cause an overflow in u64, which would take over a year.

kernel/src/time.rs
```diff
@@ -108,7 +112,8 @@ pub fn now() -> Duration {
 }
 
 pub fn from_clock_cycles(cycles: u64) -> Duration {
-    let now = 1_000_000_000 * cycles / ClockImpl::hz();
+    // Use u128 to avoid multiplication overflow when cycles is large
+    let now = (cycles as u128 * 1_000_000_000u128 / ClockImpl::hz() as u128) as u64;
```
xuchang-vivo: Ditto

han-jiang277: panic is here

han-jiang277:
build_prs

han-jiang277:
build_prs

kernel/src/time.rs
```diff
@@ -65,10 +64,8 @@ impl Tick {
 
     pub fn now() -> Self {
         debug_assert_eq!(ClockImpl::hz() % TICKS_PER_SECOND as u64, 0);
-        Self(
-            (ClockImpl::estimate_current_cycles() * TICKS_PER_SECOND as u64 / ClockImpl::hz())
-                as usize,
-        )
+        let cycles_per_tick = ClockImpl::hz() / TICKS_PER_SECOND as u64;
```
xuchang-vivo, line 67: Integer division consumes about ten times more CPU cycles than multiplication

xuchang-vivo: The number of bits in a counter generally does not exceed fifty or so, and I believe the previous implementation was fine.

han-jiang277: `let cycles_per_tick = ClockImpl::hz() / TICKS_PER_SECOND as u64;` this maybe const

xuchang-vivo: Right!

xuchang-vivo (approved):

han-jiang277:
build_prs

han-jiang277:
build_prs

zhaowenbo-vivo (approved):

PR #378, opened 2026-03-23
fix: SMP priority inheritance race condition in mutex lock chain

### LINKED ISSUE #377: test_blocking_mutex_chain_basic panic in qemu_virt64_aarch64.debug.swi
https://github.com/vivoblueos/kernel/actions/runs/23403379438/job/68078231198

```
[ RUN      ] test_blocking_mutex_chain_basic
panicked at ../../kernel/kernel/src/sync/mutex.rs:732:17:
assertion `left == right` failed
  left: 3
 right: 0
Oops: assertion `left == right` failed
```

han-jiang277:
## Summary

Fixes a race condition in SMP environments where priority inheritance (PI) lock chain traversal in `inner_pend_for` uses stale priority values, causing `test_blocking_mutex_chain_basic` to fail intermittently.

## Root Cause

In SMP environments, two threads can execute `inner_pend_for` simultaneously:
1. Thread A captures priority value (e.g., 6)
2. Thread A releases its lock, acquires the scanned thread's lock
3. During this window, Thread B promotes Thread A's priority (from 6 to 0)
4. Thread A uses the stale value (6) for comparison, incorrectly skipping priority promotions

## Fix https://github.com/vivoblueos/kernel/issues/377

- Re-read `current_priority` after acquiring `scanning_thread`'s lock to get the most up-to-date value
- Use conditional promotion instead of forced promotion to avoid overwriting already-promoted priorities
- Continue traversing the lock chain even if `adjust_wait_queue_position_by` fails

## Test Plan

- [x] `test_blocking_mutex_chain_basic` passes 50000 iterations on qemu_virt64_aarch64.debug.swi (SMP 8-core)
- [x] All mutex tests pass on qemu_virt64_aarch64.debug.swi
- [x] All kernel tests pass on qemu_mps2_an385.release


han-jiang277:
build_prs

han-jiang277:
build_prs

kernel/src/time/timer.rs
```diff
@@ -380,7 +380,7 @@ mod tests {
             tm.mode = TimerMode::Repeat(r);
             tm.callback = callback;
             iou = add_soft_timer(&mut tm).unwrap();
-            scheduler::suspend_me_until::<()>(base_deadline.add(Tick(21)), None);
+            scheduler::suspend_me_until::<()>(base_deadline.add(Tick(30)), None);
```
nooopslol: Will setting Tick(30) alleviate the SMP priority inheritance contention issue? whether more  callbacks will bring any negative impacts?

kernel/src/time/timer.rs
```diff
@@ -380,7 +380,7 @@ mod tests {
...
             iou = remove_soft_timer(iou).unwrap();
```
nooopslol, line 385: add_soft_timer   
scheduler::suspend_me_until   
remove_soft_timer

Whether remove_soft_timer may not execute in case task cancel or timeout exception?  it might be necessary to add something like a timer guard to ensure automatic remove_soft_timer after drop.

han-jiang277: timer is create in stack now , It will drop automatically.

kernel/src/time/timer.rs
han-jiang277: this is not SMP priority inheritance contention issue, will be rolled back.

han-jiang277:
build_prs

han-jiang277:
build_prs

han-jiang277:
build_prs

zhaowenbo-vivo (approved):

han-jiang277:
build_prs

PR #382, opened 2026-03-25
Add NUM_IRQS in BSP

xuchang-vivo:
The amount of IRQs are different between different Chips.

These info should be provided on BSP.

See https://docs.zephyrproject.org/latest/hardware/porting/soc_porting.html


xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

PR #392, opened 2026-03-31
fix(time): eliminate race condition in test_oneshot_swtm

han-jiang277:
Replace time-based waiting with counter polling to avoid race condition caused by QEMU RISC-V tick drift under high load.

Fixes #389


han-jiang277:
build_prs

yyang0814 (approved):

han-jiang277:
build_prs

han-jiang277:
build_prs

PR #395, opened 2026-04-07
Add early printk supports

xuchang-vivo:
"Provides a minimal SYSLOG output facility that can be used during the very early boot phase or when the system is in a down state, before the full SYSLOG subsystem or scheduler becomes available. And sends the resulting characters directly to the low-level output device using corrsponding registers.  It is primarily intended for debugging or diagnostic messages in contexts where interrupts may be disabled and locking is not yet functional."

Inspired by https://github.com/apache/nuttx/blob/master/drivers/syslog/syslog_early.c

### DIFF
--- a/hal/src/lib.rs
+++ b/hal/src/lib.rs
@@ -91,9 +91,7 @@ pub trait HasLineStatusReg {
     ///
     /// This method is typically used to ensure the bus is idle before starting
     /// a new transmission, preventing data corruption or transmission conflicts.
-    fn is_bus_busy(&self) -> bool {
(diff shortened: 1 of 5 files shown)


xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

nooopslol (approved):

PR #400, opened 2026-04-16
Fix problem in start of aarch64

jinwjinl:
The 4k stack space was invalid because it was created after the function call. Furthermore, the sp value was changed before the "enable_mmu" function was called, but was not changed back afterward.


jinwjinl:
build_prs

jinwjinl:
build_prs

han-jiang277 (approved):

PR #406, opened 2026-04-28
Add async tasklet waker

xuchang-vivo:
## Summary
- add a per-tasklet RawWaker for the kernel async executor
- requeue pending tasklets only when their waker fires
- add a yield_now future and unit coverage

## Tasklet State Model
The async executor tracks two things for each tasklet: whether it is currently linked in `ASYNC_WORK_QUEUE`, and the tasklet scheduling state.

```mermaid
stateDiagram-v2
    [*] --> IDLE: created

    IDLE --> QUEUED: wake and enqueue
    QUEUED --> POLLING: pop from queue and poll

    POLLING --> IDLE: pending and not woken
    POLLING --> WOKEN: woken during poll
    WOKEN --> QUEUED: requeue after pending

    POLLING --> COMPLETED: ready
    COMPLETED --> [*]: no Arc refs remain
```

Queue ownership by state:

| State | In async work queue? | Meaning |
| --- | --- | --- |
| `IDLE` | No | The tasklet is pending but not scheduled. A later waker call can enqueue it. |
| `QUEUED` | Yes | The tasklet is already scheduled. More wake calls are coalesced. |
| `POLLING` | No | The poller popped it from the queue and is currently polling its future. |
| `WOKEN` | No | The tasklet was woken while it was being polled; it will be requeued after `Pending`. |
| `COMPLETED` | No | The future returned `Ready` and will not be scheduled again. |

`pop_front()` does not destroy the tasklet. It transfers the list-owned `Arc<Tasklet>` into the local `task` variable in `poll_inner()`. If the future stores `cx.waker()`, that waker also owns an `Arc<Tasklet>`, so future events can schedule the same tasklet again.

## Validation
- ninja -C out/qemu_riscv64.release kernel/kernel:check_kernel_by_clippy

### DIFF
--- a/kernel/src/asynk/mod.rs
+++ b/kernel/src/asynk/mod.rs
@@ -32,7 +32,7 @@ use core::{
     mem::MaybeUninit,
     pin::Pin,
     sync::atomic::{AtomicUsize, Ordering},
-    task::{Context, Poll, Waker},
+    task::{Context, Poll, RawWaker, RawWakerVTable, Waker},
 };
 
 impl_simple_intrusive_adapter!(TaskletNode, Tasklet, node);
@@ -43,6 +43,7 @@ pub struct Tasklet {
     lock: ISpinLock<Tasklet, TaskletLock>,
     future: Pin<Box<dyn Future<Output = ()>>>,
     blocked: Option<ThreadNode>,
+    state: AtomicUsize,
 }
 
 impl Tasklet {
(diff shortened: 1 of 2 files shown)


xuchang-vivo:
build_prs

han-jiang277 (commented):
pls add more test cases

han-jiang277:
pls add async design in book.

han-jiang277 (approved):

PR #411, opened 2026-05-10
[codex] Slim ARM kernel arch with bluekernel_arch

xuchang-vivo:
## Summary

This PR slims down the ARM kernel architecture layer by moving Cortex-M-only building blocks into `bluekernel_arch`.

The kernel side now keeps the OS-specific exception and scheduler glue, while architecture-owned pieces such as NVIC/SCB/SysTick/MPU wrappers, xPSR and exception-frame helpers, HardFault decoding, IRQ registry logic, saved `Context`, local IRQ helpers, and the fixed Cortex-M exception vector layout live under `kernel/arch`.

## Why

The main goal is to make `kernel/src/arch/arm` smaller and more focused. Cortex-M register layouts, exception frames, vector-table slots, and raw architectural helpers do not depend on kernel policy, so keeping them in the kernel makes the kernel layer heavier than necessary and mixes two different responsibilities.

Putting those pieces into `bluekernel_arch` gives BlueOS a local, controllable architecture crate. That is useful for maintenance because BlueOS uses GN: rustflags are not transparently forwarded into external Rust dependencies. For example, using the upstream `cortex-m` crate's FPU register support may require passing feature/cfg such as `has_fpu`; with the current GN integration that can force changes around an external dependency boundary. Maintaining a small in-tree low-level architecture crate avoids that friction, and with current AI-assisted development the cost of writing and maintaining these thin register/assembly wrappers is manageable.

## Changes

- Re-export safe Cortex-M policy APIs from `bluekernel_arch` and remove duplicate kernel-side implementations.
- Move Cortex-M HardFault handling and IRQ registry code into `bluekernel_arch`.
- Move the saved Cortex-M `Context` type into `bluekernel_arch`, including hard-float ABI support.
- Define the Cortex-M exception vector slots in `bluekernel_arch` with `global_asm!`; the kernel now only provides the OS-specific handlers.
- Add linker defaults so NMI, MemManage, BusFault, and UsageFault can fall back to the HardFault handler when the kernel does not override them.
- Keep GOT sections inside the copied data range so optimized Rust code that emits GOT references still gets initialized correctly.
- Replace the old `IsrContext` alias usage with `ExceptionStackFrame`.
- Remove the CMSIS validation target's external `cortex-m` crate dependency and route its NVIC test shims through `bluekernel_arch`.

## Validation

- Formatted touched Rust files with the fixed BlueOS toolchain rustfmt:
  `/Users/xuchang/workspace/bluekernel_toolchain/usr/local/bin/rustfmt ...`
- Checked touched GN formatting:
  `gn format --dry-run adapter/cmsis/BUILD.gn`
- Checked CMSIS validation Rust formatting:
  `/Users/xuchang/workspace/bluekernel_toolchain/usr/local/bin/rustfmt --check adapter/cmsis/src/validation.rs`
- `ninja -C out/qemu_mps2_an385.release kernel/arch:bluekernel_arch kernel/kernel:blueos`
- `ninja -C out/qemu_mps2_an385.release kernel/adapter/cmsis:adapter_validation`
- `ninja -C out/qemu_mps2_an385.release kernel/adapter/cmsis:run_adapter_validation`
- `ninja -C out/qemu_mps2_an385.release check_all`

Note: `run_adapter_validation` still prints the existing `TC_osEventFlagsNew_1 FAILED` CMSIS validation case, but the ninja target exits successfully as before.


xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

PR #414, opened 2026-05-11
fix(linker): place GOT sections within .data to ensure proper init

xuchang-vivo:
## Summary

- Include `.got`, `.got.*`, `.igot.*`, `.got.plt`, and `.igot.plt` sections inside the `.data` output section in the Cortex-M linker script
- Without this, `.got` becomes an orphan section placed after `__data_end`, so the startup copy table does not copy it from flash to RAM, leaving its entries zeroed at runtime
- This affects bare-metal Rust code where the compiler may emit PC-relative loads referencing .got entries

### DIFF
--- a/arch/arm/cortex_m/link.x
+++ b/arch/arm/cortex_m/link.x
@@ -176,6 +176,21 @@ SECTIONS
     __stop___llvm_prf_data = .;
 
     KEEP(*(.jcr*))
+
+    /*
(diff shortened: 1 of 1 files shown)


xuchang-vivo:
build_prs

han-jiang277 (approved):

zhaowenbo-vivo (approved):

PR #417, opened 2026-05-13
Add coredump

zhaowenbo-vivo:
## Summary

Implement a complete on-device coredump subsystem that generates ELF ET_CORE
files on fatal faults, supporting all 4 kernel architectures (RISC-V 32/64,
AArch64, Cortex-M).

### Key components

- **ELF builder** (`elf.rs`): Hand-written ELF32/ELF64 `#[repr(C)]` structs,
  5-phase builder (header → PHDRs → notes → segments → backfill)
- **Register capture** (`arch/`): Per-architecture register dumps following
  Linux `user_regs_struct` layout
  - RISC-V: 32 × usize (pc, ra, sp, gp, tp, t0-t6, fp, a0-a7, s1-s11)
  - AArch64: 34 × 8 bytes (x0-x30, sp, pc, pstate)
  - Cortex-M: 18 × 4 bytes (r0-r15, cpsr, orig_r0)
- **Signal mapping** (`signal.rs`): Maps arch-specific exception codes to POSIX
  signals (SIGSEGV, SIGBUS, SIGILL, SIGFPE, SIGTRAP, SIGABRT)
- **Memory regions** (`regions.rs`): Collects BSS, system stack, and thread
  stacks via `GlobalQueueVisitor`
- **Three backends** (`backend.rs`):
  - `MemoryBackend`: Static buffer for post-mortem retrieval
  - `FileBackend`: VFS file output (`/tmp/blueos.core.<pid>`)
  - `LoggingBackend`: Hex-encoded serial/semihosting output
- **Integration test**: Runs in QEMU with `.checker` assertion validation
- **Parser tool**: `coredump_parser.py` reassembles hex output → `.core` ELF
- **Usage docs**: `README.md` with backend selection, Kconfig options, and tool usage

### Kconfig options

- `ENABLE_COREDUMP` (bool) — global switch
- `COREDUMP_BUF_SIZE` (int) — buffer size per platform (64KB Cortex-M,
  256KB 32-bit, 4MB 64-bit)
- `COREDUMP_MEM_BACKEND` (bool) — MemoryBackend selection

### Code quality

- `#[inline(never)]` on all exception entry points (`capture_regs`,
  `capture_current_regs`, `dump()`)
- `Debug`/`Clone` derives on all public types
- `// SAFETY:` comments on every `unsafe` block
- No `unwrap()` in core logic

## Test plan

- [x] Compile on all 4 architectures (riscv64, riscv32, aarch64, cortex-m)
- [x] Integration test passes in QEMU (debug config)
- [x] All 27 board/config defconfigs updated with `CONFIG_ENABLE_COREDUMP=y`
- [x] Release/coverage builds verified

EOF
)"


kernel/src/Kconfig
```diff
@@ -65,3 +65,5 @@ menu "os adapter configuration"
 
     endchoice
 endmenu # os adapter configuration
+
+rsource "coredump/Kconfig"
```
xuchang-vivo, line 69: Plz also put the path on the `_kconfig_files`, otherwise the build system can not trace the changes of coredump/Kconfig

kernel/tests/coredump_test.rs
xuchang-vivo, line 1: Missing assertion condition, how to determine the failure condition

kernel/src/coredump/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+/// Must be large enough to hold the ELF header, program headers, notes,
+/// and all segment data (BSS, stacks). On 64-bit platforms where BSS can
+/// be 2+ MB and system stacks 512 KB, 4 MB is required. On 32-bit
+/// platforms 256 KB suffices.
+///
+/// Using a static buffer instead of a stack allocation avoids stack
+/// overflow since thread stacks (64 KB) are much smaller than the
+/// coredump buffer.
+#[link_section = ".coredump_bss"]
+static mut ELF_BUF: [u8; COREDUMP_BUF_SIZE] = [0u8; COREDUMP_BUF_SIZE];
```
han-jiang277, line 49: Why use the BSS segment for memory space that is never used at runtime?

han-jiang277: coredump may not trigger by signal when run in RTOS mode.

kernel/src/coredump/signal.rs
xuchang-vivo, line 1: Put this functionailty into arch might be a better choice

kernel/src/coredump/regions.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    match mode {
+        DumpMode::Current => {
+            collect_current_stack(&mut col);
+        }
+        DumpMode::Threads => {
+            collect_all_stacks(&mut col);
+        }
+        DumpMode::Full => {
+            collect_all_stacks(&mut col);
```
xuchang-vivo, line 166: When the mode is ALL, will the stack of the current thread definitely be recorded?

kernel/src/coredump/backend.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn write(&mut self, buf: &[u8]) {
+        // Output in small hex-encoded lines via console
+        const LINE_MAX: usize = 64; // 32 bytes per line → 64 hex chars
+        let mut hex_buf = [0u8; LINE_MAX * 2 + 1]; // +1 for newline
+        for chunk in buf.chunks(LINE_MAX) {
+            let n = hex_encode(chunk, &mut hex_buf);
+            hex_buf[n] = b'\n';
+            // Use semihosting eprint which is always available
+            let s = unsafe { core::str::from_utf8_unchecked(&hex_buf[..n + 1]) };
+            semihosting::eprint!("{}", s);
```
xuchang-vivo, line 138: Semihosting is only used under qemu test, and it is not suitable for use here.

kernel/src/coredump/backend.rs
```diff
@@ -0,0 +1,193 @@
...
+impl CoredumpBackend for LoggingBackend {
+    fn write(&mut self, buf: &[u8]) {
+        // Output in small hex-encoded lines via console
+        const LINE_MAX: usize = 64; // 32 bytes per line → 64 hex chars
+        let mut hex_buf = [0u8; LINE_MAX * 2 + 1]; // +1 for newline
+        for chunk in buf.chunks(LINE_MAX) {
+            let n = hex_encode(chunk, &mut hex_buf);
+            hex_buf[n] = b'\n';
+            // Use semihosting eprint which is always available
+            let s = unsafe { core::str::from_utf8_unchecked(&hex_buf[..n + 1]) };
```
xuchang-vivo, line 137: Calling `from_utf8_unchecked` on hex_buf will result in undefined behavior.

kernel/src/coredump/backend.rs
```diff
@@ -0,0 +1,193 @@
...
+            semihosting::eprint!("{}", s);
+        }
+        self.chunks += 1;
+    }
+
+    fn finalize(&mut self, mode: DumpMode) {
+        // Emit a trailer line with metadata
+        let size_kb = self.chunks * 32 / 1024;
+        semihosting::eprint!(
```
xuchang-vivo, line 146: Dtto

xuchang-vivo (changes requested):

kernel/src/coredump/backend.rs
xuchang-vivo: A better approach is to use `write_data8`, which is part of `Has8bitDataReg`. You can refer to the implementation of `kearly_printk` for reference.

xuchang-vivo: But in fact, using `kearly_print` is not a good method, as it will mix the output of these two channels together. A separate channel should be provided for it. I suggest leaving this blank and writing a todo or fix note. We can come back to fix it once I find an elegant way to connect to the second serial port channel.

PR #422, opened 2026-06-05
Tlsf align overhead reduction

han-jiang277:
```
Current (absorb padding) — 4 KiB-aligned, 4 KiB payload, GRANULARITY = 32 (64-bit)
 block_start (G-aligned)                         aligned_payload
 │                                               │
 ▼                                               ▼
 ┌──────┬───────── dead padding ─────────┬──────┬──────── PAYLOAD ────────┐
 │BlkHdr│        ~4080 B (lost)          │ Pad  │         4096 B           │
 └──────┴───────────────────────────────┴──────┴─────────────────────────┘
 └───────────────── used block ≈ 8192 B (retained) ──────────────────────┘


Proposed (reclaim leading gap)
 block_start          U_start = aligned_payload - G      aligned_payload
 │                    │                                  │
 ▼                    ▼                                  ▼
 ┌──── FREE  L ───────┬──────┬──────┬──── PAYLOAD ───────┬─── FREE T ───┐
 │  ~4064 B (reused)  │BlkHdr│ Pad  │      4096 B         │  remainder   │
 └────────────────────┴──────┴──────┴────────────────────┴──────────────┘
 phys chain:  prev ← L ← U(used) ← T ← next      (all prev_phys fixed up)
 retained for this alloc ≈ 4128 B   (was ≈ 8192 B)
```

### DIFF
--- a/kernel/src/allocator/block.rs
+++ b/kernel/src/allocator/block.rs
@@ -15,6 +15,7 @@
 // This code is based on https://github.com/yvt/rlsf/blob/main/crates/rlsf/src/tlsf.rs.
 // Copyright 2021 yvt
 // SPDX-LICENSE: MIT
+// Divergence from upstream: overhead model updated to reflect leading-gap reclamation (align >= GRANULARITY reduces retained footprint to size + GRANULARITY).
 
 use core::{
     alloc::Layout,
--- a/kernel/src/allocator/tlsf/mod.rs
+++ b/kernel/src/allocator/tlsf/mod.rs
@@ -1619,27 +1619,33 @@ mod tests {
     use super::*;
     use blueos_test_macro::test;
 
-    // On 64-bit: GRANULARITY=32. Use u8 bitmaps for small tests.
-    type TestTlsf<'a> = Tlsf<'a, u8, u8, 8, 8>;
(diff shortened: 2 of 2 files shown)


han-jiang277:
build_prs

han-jiang277:
build_prs

han-jiang277:
Two reasons block.rs doesn't need code changes:

1. get_overhead_and_size — search size stays the same by design

This function computes the search size — how large a free block to look for. After reclamation, we still need to find a block of that size because the gap must physically exist before we can carve it. The search size is a mathematical floor, not the retained size. Changing it would break the allocator.

2. used_block_hdr_for_allocation — header location protocol unchanged

For align >= GRANULARITY, it reads the UsedBlockPad pointer stored immediately before the payload. After reclamation:

U_start = aligned_payload - GRANULARITY
UsedBlockPad sits at aligned_payload - sizeof(UsedBlockPad) = aligned_payload - GRANULARITY/2
The pad stores a pointer back to U_start
This is identical to the non-reclamation path — the pad is always placed just before the payload, and always points to the header. The relative geometry between payload, pad, and header is unchanged; only the header's absolute position in the pool moved forward. free/deallocate just follows the pad pointer and finds the right header regardless.

So block.rs needed only a comment update (provenance note), not a code change.

yyang0814 (approved):

xuchang-vivo (approved):
Look good

PR #423, opened 2026-06-12
Draft: Split arch from kernel

han-jiang277:


han-jiang277:
build_prs

blueos_boards/src/qemu_virt64_aarch64/config.rs
zhaowenbo-vivo, line 1: 走kcofig

blueos_allocator/src/llff/heap.rs
zhaowenbo-vivo, line 1: blueos前缀去掉

PR #427, opened 2026-06-17
Add async channel

xuchang-vivo:
## Summary

Add a SPSC (Single-Producer Single-Consumer) asynchronous channel implementation for the BlueOS kernel async executor.

## Changes

- **channel.rs**: Fixed-capacity lock-free ring-buffer channel with:
  - Async `send`/`recv` via `SendFuture`/`RecvFuture`
  - Blocking `send_blocking`/`recv_blocking` via `atomic_wait`
  - Disconnect detection on `Sender`/`Receiver` drop
  - Explicit `close()` from either end
  - Capacity must be a power of two (`N > 1`)
  - Unit tests covering try_send/recv, blocking, async, disconnect, wrap-around

- **mod.rs**: Added `pub(crate) mod channel;` declaration

## Files Changed

| File | Lines |
|------|-------|
| `kernel/src/asynk/channel.rs` | +657 (new) |
| `kernel/src/asynk/mod.rs` | +1 |

### DIFF
--- a/kernel/src/asynk/channel.rs
+++ b/kernel/src/asynk/channel.rs
@@ -408,13 +408,6 @@ impl<T, const N: usize> Drop for Receiver<T, N> {
     }
 }
 
-impl<T, const N: usize> Clone for Receiver<T, N> {
-    fn clone(&self) -> Self {
-        // SPSC: cloning receiver breaks the protocol. Panic to signal misuse.
-        panic!("Receiver::clone() called on an SPSC channel (only one consumer allowed)");
-    }
-}
-
 // ---------------------------------------------------------------------------
 // SendFuture
 // ---------------------------------------------------------------------------
--- a/kernel/src/asynk/channel.rs
+++ b/kernel/src/asynk/channel.rs
@@ -122,7 +122,16 @@ impl<T, const N: usize> core::fmt::Debug for ChanInner<T, N> {
 }
 
 unsafe impl<T: Send, const N: usize> Send for ChanInner<T, N> {}
-unsafe impl<T: Send + Sync, const N: usize> Sync for ChanInner<T, N> {}
+
+// SAFETY: `ChanInner` is shared by exactly one `Sender` and one `Receiver`.
+// The producer is the only writer of `head` and the only initializer of empty
+// slots; the consumer is the only writer of `tail` and the only reader that
+// moves initialized values out. The release/acquire ordering on `head` and
+// `tail` publishes slot initialization before receive and slot reclamation
+// before reuse. No shared references to `T` are ever exposed, so `T: Sync` is
+// not required; `T: Send` is enough because values are transferred by ownership
+// between execution contexts.
+unsafe impl<T: Send, const N: usize> Sync for ChanInner<T, N> {}
 
 impl<T, const N: usize> ChanInner<T, N> {
     const fn new() -> Self {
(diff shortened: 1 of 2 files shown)


xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

han-jiang277:
## Code Review: `kernel/src/asynk/channel.rs`

| # | Severity | Line | Summary |
|---|----------|------|---------|
| 1 | **Critical** | 287 | `Sender` derives `Clone` but channel is SPSC — two concurrent `try_send` calls race on `head`, corrupting the ring buffer and exposing uninitialized memory |
| 2 | **Critical** | 131 | `unsafe impl Sync` for `ChanInner` unsound — bound `T: Send` should be `T: Sync` (wraps `UnsafeCell<MaybeUninit<T>>`) |
| 3 | **High** | 448 | `SendFuture::poll` lost wakeup — no re-check of `try_send` after storing waker; future parks permanently |
| 4 | **High** | 437 | `SendFuture::poll` leaks item on disconnect — sets `done=true`, returns `Err(SendError)`, but `Drop` skips `assume_init_drop` |
| 5 | **Medium** | 179 | `try_send` TOCTOU race — `is_disconnected()` check and slot write are not atomic; item can be written to a dead channel |
| 6 | **Medium** | 446 | `SendFuture::poll` missing disconnect re-check after waker store — disconnect races with waker registration |
| 7 | **Low** | 409 | `Receiver::clone` panics at runtime — should not implement `Clone` at all (compile-time safety) |

### Details

**Finding #1 (Critical):** `Sender` derives `Clone` (line 287) but `try_send` uses a non-atomic read-modify-write on `head`:
```rust
let head = self.head.load(Ordering::Relaxed);  // two senders read same head
unsafe { (*self.buf[Self::mask(head)].val.get()).write(val); }  // both write same slot
self.head.store(head.wrapping_add(1), Ordering::Release);  // both store head+1
```
Two cloned senders both read the same `head`, both write to the same slot (overwriting), and the consumer reads uninitialized memory for the phantom second item. The `Receiver` correctly panics on clone (line 412) — `Sender` should do the same.

**Finding #2 (Critical):** `ChanInner` contains `[Slot<T>; N]` where `Slot<T>` wraps `UnsafeCell<MaybeUninit<T>>`. `UnsafeCell` is `Sync` only when `T: Sync`, but the unsafe impl only requires `T: Send`. Fix: add `T: Sync` bound.

**Finding #3 (High):** `SendFuture::poll` stores the waker then returns `Pending` without re-checking `try_send`:
```rust
Err(val) => {
    this.item.write(val);
    *this.inner.send_waker.lock() = Some(cx.waker().clone());
    Poll::Pending  // <-- no re-check of try_send after storing waker
}
```
The `RecvFuture::poll` (line 486) correctly re-checks `try_recv` after storing its waker. Add the same pattern to `SendFuture`.

**Finding #4 (High):** When `is_disconnected()` is true at line 436, `done` is set to `true` and `Err(SendError)` is returned. The value in `this.item` (initialized at line 308) is never consumed by `assume_init_read()` (line 440 is unreachable), and `Drop` skips `assume_init_drop()` because `done==true`. The value is leaked.

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

han-jiang277 (approved):

PR #441, opened 2026-07-27
Support Loading Fixed-Address ET_EXEC Images

YinhuaChen-vivo:
## Summary

This change extends the ELF loader to support statically linked `ET_EXEC` images that must be loaded at their link-time virtual addresses. This is required by targets such as ESP32-C3, where the test payload is placed in a board-provided executable memory region instead of allocator-backed memory.

The existing position-independent `ET_DYN` path remains allocator-backed and continues to apply supported dynamic relocations.

```text
                         +-------------------+
ELF buffer ------------>| parse ELF header  |
                         +---------+---------+
                                   |
                     +-------------+-------------+
                     |                           |
                 e_type=ET_DYN                e_type=ET_EXEC
                     |                           |
                     v                           v
             +---------------+           +------------------+
             | Allocated mode|           | Fixed mode       |
             +-------+-------+           +--------+---------+
                     |                            |
             allocate backing store       validate authorized
                     |                     address ranges
                     v                            |
             copy PT_LOAD segments               v
                     |                     copy PT_LOAD segments
                     v                     to linked addresses
             apply relocations                    |
                     |                            v
                     +-----------> validate executable entry
                                                  |
                                                  v
                                            return entry point
```

## What Changed

- Dispatch ELF loading by `e_type` and reject unsupported ELF types.
- Require allocator-backed mapping for `ET_DYN` images.
- Require fixed-address mapping for `ET_EXEC` images.
- Add public `MemoryRegion` and `MemoryPermissions` types so callers can explicitly authorize fixed address ranges.
- Validate every fixed-address write before copying segment contents.
- Validate that the final `ET_EXEC` entry point belongs to an executable region.
- Preserve the existing allocation and RISC-V relative relocation flow for `ET_DYN` images.
- Add an `ET_EXEC` test payload whose `.text`, `.data`, and `.bss` sections are linked into the configured test region.
- Generate the test linker script from board-provided region, origin, length, and permission values.
- Select the PIC or fixed-address test payload automatically according to whether `loader_test_relocation` is defined.
- Use the ESP32-specific QEMU runner when `qemu_use_esp32_loader` is enabled.
- Reuse the selected loader test application when generating malformed ELF inputs.

## Mapping Model

```text
Allocated mode (`MemoryMapper::new(None)`)

ELF virtual range                Allocated storage
+----------------------+         +----------------------+
| vaddr ... vaddr+size | ------> | base ... base+size   |
+----------------------+ offset  +----------------------+


Fixed mode (`MemoryMapper::new(Some(regions))`)

ELF virtual address             Authorized physical mapping
+----------------------+         +----------------------+
| linked address       | ======> | same address         |
+----------------------+ direct  +----------------------+
                                 must fit one declared region
```

Fixed mappings are intentionally explicit. Constructing a `MemoryRegion` is unsafe because the caller must guarantee that the advertised range is mapped, has the declared permissions, remains valid while the loader uses it, and does not receive conflicting accesses.

```text
PT_LOAD destination span
          |
          v
+---------------------------+
| inside one valid region?  |-- no --> reject before copying
+-------------+-------------+
              | yes
              v
       copy segment bytes

ELF entry point
          |
          v
+---------------------------+
| inside executable region? |-- no --> reject image
+-------------+-------------+
              | yes
              v
       return entry address
```

## Test Coverage

The integration test matrix now covers both loader modes:

| Scenario | Expected result |
| --- | --- |
| `ET_DYN` with allocated mapping | Load, relocate, and execute successfully |
| `ET_EXEC` with fixed mapping | Load at linked addresses and execute successfully |
| `ET_EXEC` with allocated mapping | Reject the mapping mode |
| `ET_EXEC` outside the authorized range | Reject before modifying the destination memory |
| `ET_EXEC` entry in a non-executable region | Reject the entry point |
| Invalid ELF magic | Reject |
| Invalid entry point | Reject |
| Invalid segment size | Reject |

The fixed-address payload also verifies initialized data, zero-initialized data, writable memory, and executable code by returning a deterministic value after execution.

## Compatibility

This changes `MemoryMapper::new()` to accept an optional fixed-region list:

```rust
// Existing allocator-backed behavior for ET_DYN.
let mapper = MemoryMapper::new(None);

// Fixed-address behavior for ET_EXEC.
let mapper = MemoryMapper::new(Some(&regions));
```

Boards without `loader_test_relocation` continue to build and run the existing PIC loader test application. Boards that define it build the new `ET_EXEC` payload and pass the same memory contract to the integration test at compile time.


YinhuaChen-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/441 https://github.com/vivoblueos/build/pull/73

YinhuaChen-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/441 https://github.com/vivoblueos/build/pull/73

loader/tests/integration_test/BUILD.gn
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2025 vivo Mobile Communication Co., Ltd.
+# Copyright (c) 2026 vivo Mobile Communication Co., Ltd.
```
xuchang-vivo: It is not necessary to change this line

loader/tests/integration_test/src/test.rs
```diff
@@ -101,6 +101,93 @@ mod test_everything {
     }
 }
 
+#[cfg(loader_test_exec)]
+mod test_exec {
+    use super::*;
+    use blueos_test_macro::test;
+    use core::ffi::c_char;
+
+    const RTC_FAST_START: usize = 0x5000_0000;
```
xuchang-vivo: This is board-specific. What would happen if I enable `loader_test_exec` on a non-ESP32C3 platform?

loader/tests/integration_test/src/test.rs
```diff
@@ -101,6 +101,93 @@ mod test_everything {
...
+            loader::MemoryPermissions::READ.bitor(loader::MemoryPermissions::WRITE),
+        )
+    }];
+
+    extern "C" {
+        static EXEC_ELF_PATH: *const c_char;
+    }
+
+    #[test]
+    fn test_load_exec_elf_and_run() {
```
xuchang-vivo: It seems unnecessary to specifically distinguish between `load_test_pic` and `load_test_elf`. The `new` and `new_fix` functions can be unified into a single interface, differentiating object creation based on whether the argument is None. Other test cases appear to differ only in their paths.

And this path can serve as a unified entry point; what needs to be configured is `relocate`. Please refer to Zephyr's relocate region configuration: https://docs.zephyrproject.org/latest/kernel/code-relocation.html

loader/tests/inputs/BUILD.gn
```diff
@@ -14,8 +14,12 @@
 
 group("inputs") {
   testonly = true
-  deps = [
-    "malware:bomb_path",
-    "no_std_app:everything_elf_path",
-  ]
+  if (board == "seeed_xiao_esp32c3") {
+    deps = [ "exec_app:exec_elf_path" ]
```
xuchang-vivo: Can this path be unified into a single input?

If I need to add ten boards in the future, what will this `if statement` look like? like seeed_xiao_esp32c5, seeed_xiao_esp32c5, esp32c3-devkim, etc

loader/tests/inputs/no_std_app/BUILD.gn
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2025 vivo Mobile Communication Co., Ltd.
```
xuchang-vivo, line 1: dotto

xuchang-vivo (changes requested):

loader/tests/inputs/no_std_app/BUILD.gn
YinhuaChen-vivo: done

loader/tests/integration_test/BUILD.gn
YinhuaChen-vivo: done

loader/tests/integration_test/src/test.rs
YinhuaChen-vivo: > The new and new_fix functions can be unified into a single interface, differentiating object creation based on whether the argument is None. 

done

YinhuaChen-vivo: > Other test cases appear to differ only in their paths.
> And this path can serve as a unified entry point;

done.

Already unify PIC_ELF_PATH and EXEC_ELF_PATH into LOADER_TEST_ELF_PATH.

Other paths such as INVALID_MAGIC_ELF_PATH serve tests intended for different purposes. Unifying these paths is inappropriate if corresponding tests are not suitable to combine.

YinhuaChen-vivo:
build_prs https://github.com/vivoblueos/kernel/pull/441 https://github.com/vivoblueos/build/pull/73

loader/tests/integration_test/src/test.rs
YinhuaChen-vivo: > It seems unnecessary to specifically distinguish between load_test_pic and load_test_elf.

done. Already unified.

YinhuaChen-vivo: > what needs to be configured is relocate. Please refer to Zephyr's relocate region configuration: https://docs.zephyrproject.org/latest/kernel/code-relocation.html

done.

Following Zephyr's design philosophy, an exec_app linker script configurable via build/boards/\<board\>.gni has been added.

loader/tests/integration_test/src/test.rs
YinhuaChen-vivo: done.

In the updated code, board-specific parameters are defined in build/boards/\<board\>.gni.

loader/tests/inputs/BUILD.gn
YinhuaChen-vivo: done.

No 'board == "seeed_xiao_esp32c3"' any more. 

In the updated code, it is replaced by:
```
import("//build/boards/${board}.gni")
```

xuchang-vivo (approved):

han-jiang277 (approved):

YinhuaChen-vivo:
This PR was closed due to the accidental deletion of the forked repository.

The changes have been resubmitted under PR https://github.com/vivoblueos/kernel/pull/444.

PR #444, opened 2026-08-06
Support Loading Fixed-Address ET_EXEC Images

YinhuaChen-vivo:
## Summary

This change extends the ELF loader to support statically linked `ET_EXEC` images that must be loaded at their link-time virtual addresses. This is required by targets such as ESP32-C3, where the test payload is placed in a board-provided executable memory region instead of allocator-backed memory.

The existing position-independent `ET_DYN` path remains allocator-backed and continues to apply supported dynamic relocations.

```text
                         +-------------------+
ELF buffer ------------>| parse ELF header  |
                         +---------+---------+
                                   |
                     +-------------+-------------+
                     |                           |
                 e_type=ET_DYN                e_type=ET_EXEC
                     |                           |
                     v                           v
             +---------------+           +------------------+
             | Allocated mode|           | Fixed mode       |
             +-------+-------+           +--------+---------+
                     |                            |
             allocate backing store       validate authorized
                     |                     address ranges
                     v                            |
             copy PT_LOAD segments               v
                     |                     copy PT_LOAD segments
                     v                     to linked addresses
             apply relocations                    |
                     |                            v
                     +-----------> validate executable entry
                                                  |
                                                  v
                                            return entry point
```

## What Changed

- Dispatch ELF loading by `e_type` and reject unsupported ELF types.
- Require allocator-backed mapping for `ET_DYN` images.
- Require fixed-address mapping for `ET_EXEC` images.
- Add public `MemoryRegion` and `MemoryPermissions` types so callers can explicitly authorize fixed address ranges.
- Validate every fixed-address write before copying segment contents.
- Validate that the final `ET_EXEC` entry point belongs to an executable region.
- Preserve the existing allocation and RISC-V relative relocation flow for `ET_DYN` images.
- Add an `ET_EXEC` test payload whose `.text`, `.data`, and `.bss` sections are linked into the configured test region.
- Generate the test linker script from board-provided region, origin, length, and permission values.
- Select the PIC or fixed-address test payload automatically according to whether `loader_test_relocation` is defined.
- Use the ESP32-specific QEMU runner when `qemu_use_esp32_loader` is enabled.
- Reuse the selected loader test application when generating malformed ELF inputs.

## Mapping Model

```text
Allocated mode (`MemoryMapper::new(None)`)

ELF virtual range                Allocated storage
+----------------------+         +----------------------+
| vaddr ... vaddr+size | ------> | base ... base+size   |
+----------------------+ offset  +----------------------+


Fixed mode (`MemoryMapper::new(Some(regions))`)

ELF virtual address             Authorized physical mapping
+----------------------+         +----------------------+
| linked address       | ======> | same address         |
+----------------------+ direct  +----------------------+
                                 must fit one declared region
```

Fixed mappings are intentionally explicit. Constructing a `MemoryRegion` is unsafe because the caller must guarantee that the advertised range is mapped, has the declared permissions, remains valid while the loader uses it, and does not receive conflicting accesses.

```text
PT_LOAD destination span
          |
          v
+---------------------------+
| inside one valid region?  |-- no --> reject before copying
+-------------+-------------+
              | yes
              v
       copy segment bytes

ELF entry point
          |
          v
+---------------------------+
| inside executable region? |-- no --> reject image
+-------------+-------------+
              | yes
              v
       return entry address
```

## Test Coverage

The integration test matrix now covers both loader modes:

| Scenario | Expected result |
| --- | --- |
| `ET_DYN` with allocated mapping | Load, relocate, and execute successfully |
| `ET_EXEC` with fixed mapping | Load at linked addresses and execute successfully |
| `ET_EXEC` with allocated mapping | Reject the mapping mode |
| `ET_EXEC` outside the authorized range | Reject before modifying the destination memory |
| `ET_EXEC` entry in a non-executable region | Reject the entry point |
| Invalid ELF magic | Reject |
| Invalid entry point | Reject |
| Invalid segment size | Reject |

The fixed-address payload also verifies initialized data, zero-initialized data, writable memory, and executable code by returning a deterministic value after execution.

## Compatibility

This changes `MemoryMapper::new()` to accept an optional fixed-region list:

```rust
// Existing allocator-backed behavior for ET_DYN.
let mapper = MemoryMapper::new(None);

// Fixed-address behavior for ET_EXEC.
let mapper = MemoryMapper::new(Some(&regions));
```

Boards without `loader_test_relocation` continue to build and run the existing PIC loader test application. Boards that define it build the new `ET_EXEC` payload and pass the same memory contract to the integration test at compile time.


YinhuaChen-vivo:
This PR replaces the accidentally closed PR https://github.com/vivoblueos/kernel/pull/441.

YinhuaChen-vivo:
build_prs

YinhuaChen-vivo:
build_prs

YinhuaChen-vivo:
build_prs

xuchang-vivo (approved):

han-jiang277 (approved):

PR #448, opened 2026-08-17
add interface operating ESP32 internal flash

jinwjinl:
## Description

This branch adds ESP32-C3 on-chip Flash access and executable XIP mapping support. A reserved Flash region is exposed as `/dev/esp32-flash0`, allowing userspace loaders to erase, read, write, map, and unmap executable images without embedding ELF or image-management logic in the kernel.

### Internal Flash driver

- Adds an ESP32-C3 internal Flash driver based on the ROM SPI Flash APIs.
- Supports bounded read, aligned programming, and sector erase operations.
- Splits unaligned writes into page-sized operations and preserves unaffected bytes.
- Serializes Flash operations to prevent concurrent access.
- Validates alignment, capacity, protected ranges, and integer overflow.

### Device registration

- Registers the reserved internal Flash region as the misc device:

  ```text
  /dev/esp32-flash0
  ```

- Provides standard device operations for reading, writing, querying capacity, and synchronization.
- Adds versioned ioctl requests for:
  - erasing a Flash range;
  - mapping an executable image;
  - unmapping an executable image;
  - querying the safe DRAM boundary for loadable applications.

### XIP mapping

- Adds ESP32-C3 Flash MMU mapping support.
- Maps the reserved physical Flash region into the corresponding IROM and DROM address spaces.
- Handles 64-KiB MMU page alignment and multi-page mappings.
- Invalidates the instruction cache and executes `fence.i` after mapping changes.
- Tracks the active mapping and rejects concurrent or stale mapping operations.

### Configuration

The following parameters are defined through Kconfig:

- `ESP32_LOADABLE_REGION_BASE`
- `ESP32_LOADABLE_REGION_SIZE`
- `ESP32_FLASH_READ_CHUNK_SIZE`

The default loadable region is:

```text
Physical Flash offset: 0x00200000
Size:                  0x00100000
Mapped IROM address:   0x42200000
```

### Board integration

- Includes the ESP32-C3 ROM linker definitions required by the Flash ROM APIs.
- Initializes the internal Flash driver during ESP32-C3 boot.
- Registers `/dev/esp32-flash0` before VFS initialization.
- Keeps image metadata, ELF parsing, verification, and recovery policies in userspace.


jinwjinl:
build_prs

jinwjinl:
build_prs

jinwjinl:
build_prs

jinwjinl:
build_prs

kernel/src/drivers/flash/mod.rs
```diff
@@ -18,3 +18,21 @@
 pub mod spi_flash;
 #[cfg(enable_block)]
 pub mod spi_flash_cmd;
+
+#[cfg(soc_esp32c3)]
+mod esp32_rom;
+
+#[cfg(soc_esp32c3)]
+pub(crate) mod internal_flash;
```
xuchang-vivo, line 26: prefix `esp32`

kernel/src/drivers/flash/mod.rs
```diff
@@ -18,3 +18,21 @@
 pub mod spi_flash;
 #[cfg(enable_block)]
 pub mod spi_flash_cmd;
+
+#[cfg(soc_esp32c3)]
+mod esp32_rom;
+
+#[cfg(soc_esp32c3)]
```
xuchang-vivo, line 25: Do not use `soc_esp32c3`; instead, use a more generic gate, such as `cfg(esp32_internal_flash)`.

han-jiang277 (approved):

PR #459, opened 2026-08-26
net: derive socket timeval from target ABI

xuchang-vivo:
## Summary
- Define the socket `timeval` syscall layout directly from the BlueOS target data model.
- Use two `i32` fields on ILP32 targets and two `i64` fields on LP64 targets.
- Normalize ABI values into an internal fixed-width representation without depending on libc `time_t`, `suseconds_t`, or `timeval` definitions.
- Reject pointers, lengths, and field values that do not match the active target ABI.
- Correct the conversion from microseconds to Rust `Duration` nanoseconds.
- Add unit coverage for target layout parsing, writing, width validation, field validation, and duration conversion.

## Motivation
Socket options cross a syscall ABI boundary. Their binary layout must be determined by the supported BlueOS target ABI, not inferred from libc Rust type aliases that can change independently or disagree with the prebuilt Rust standard library.

BlueOS currently supports ILP32 and LP64 target data models:

- ILP32: `timeval { i32 tv_sec; i32 tv_usec; }`
- LP64: `timeval { i64 tv_sec; i64 tv_usec; }`

The kernel now accepts only the layout for the active target and converts it into an internal `i64` seconds/microseconds representation. This keeps the syscall ABI explicit and independent of libc implementation details.

## References
- [`setsockopt()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/setsockopt.html)
- [`getsockopt()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/getsockopt.html)
- [`timeval`](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_time.h.html)
- [RISC-V ELF psABI data models](https://riscv-non-isa.github.io/riscv-elf-psabi-doc/)

## Testing
- Added unit tests for the supported target ABI layouts and conversions.
- `rustfmt --check kernel/src/net/types.rs kernel/src/net/syscalls.rs`
- `git diff --check`
- No board build was run.

### DIFF
--- a/kernel/src/net/syscalls.rs
+++ b/kernel/src/net/syscalls.rs
@@ -30,11 +30,9 @@ use core::{
     sync::atomic::{AtomicI32, Ordering},
     time::Duration,
 };
-use libc::{size_t, timeval};
+use libc::size_t;
 use smoltcp::wire::{IpAddress, IpEndpoint};
 use spin::rwlock::RwLock;
-
-const ONE_ELEMENT: usize = 1;
 
 pub fn socket(domain: c_int, type_: c_int, protocol_: c_int) -> c_int {
     let Ok(socket_domain) = SocketDomain::try_from(domain) else {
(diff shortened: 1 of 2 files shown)


xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

han-jiang277 (approved):

PR #460, opened 2026-08-26
net: preserve socket errors across syscall boundaries

xuchang-vivo:
## Summary
- Map `ConnectionError` and `SocketError` variants to specific negative POSIX errno values.
- Preserve those errors through socket syscalls instead of collapsing every failure to `-1`.
- Preserve socket read/write errors through sockfs instead of returning the generic `code::ERROR`.
- Add unit coverage for common mappings and nested error forwarding.

## Motivation
The BlueOS network stack already distinguishes conditions such as timeout, would-block, invalid file descriptor, queue exhaustion, and address conflicts. The socket syscall and VFS boundaries previously collapsed these failures to a generic `-1`, losing the original reason before it reached libc.

For example, a timeout should leave the kernel as `-ETIMEDOUT`, allowing the libc wrapper to return `-1` with `errno = ETIMEDOUT`. Returning a bare `-1` from the kernel instead makes the libc layer interpret the error number as `1` or observe stale errno.

This complements [vivoblueos/librs#38](https://github.com/vivoblueos/librs/pull/38), which converts negative kernel errno results to the POSIX libc `-1` plus `errno` convention.

This PR only preserves reported errors. It does not implement socket timeout enforcement or change network operation behavior.

## POSIX references
- [`errno`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html)
- [`send()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/send.html)
- [`recv()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/recv.html)

## Testing
- Added unit tests for connection and socket errno mappings.
- `rustfmt --check kernel/src/net/connection_err.rs kernel/src/net/socket/socket_err.rs kernel/src/net/syscalls.rs kernel/src/vfs/sockfs.rs`
- `git diff --check`
- No board build was run.

### DIFF
--- a/kernel/src/net/connection_err.rs
+++ b/kernel/src/net/connection_err.rs
@@ -53,6 +53,22 @@ pub enum ConnectionError {
     SocketOperationError(SocketError),
 }
 
+impl ConnectionError {
+    pub fn to_errno(&self) -> i32 {
+        match self {
+            Self::SocketOperationError(error) => error.to_errno(),
+            Self::NetStackQueueFull => -libc::EAGAIN,
+            Self::PosixError(error) => error.to_errno(),
+            Self::UnsupportedSocketType(_) => -libc::EOPNOTSUPP,
+            Self::NoAvailableDynamicPort => -libc::EADDRNOTAVAIL,
+            Self::PortInUse(_) => -libc::EADDRINUSE,
+            Self::PortOutOfRange(_, _) => -libc::EINVAL,
+            Self::Timeout(_) => -libc::ETIMEDOUT,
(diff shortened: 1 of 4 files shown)


xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

han-jiang277 (approved):

Issue #456, opened 2026-08-22
net: shared OperationIPCReply can deadlock concurrent blocking I/O and shutdown


xuchang-vivo:
## Summary

A `Connection` owns one shared `OperationIPCReply` for all socket operations. Any blocking operation that returns `WouldBlock` leaves that reply in `STATE_WAITING_FOR_CONSUME` while its retry operation is held by a one-shot socket waker. A concurrent operation on the same fd cannot enqueue because `queue_and_wait()` waits for `STATE_IDLE` before enqueueing.

This is a network-layer ownership/cancellation design issue. TCP `accept()` in PR #454 exposes it clearly, but the same pattern is used by blocking TCP/UDP/ICMP send and receive operations.

## Reproducer

1. Thread A calls blocking `accept()` on a listening fd.
2. The network thread handles the request, observes `WouldBlock`, registers the recv waker, and returns without waking the shared reply.
3. Thread A remains blocked with `reply_futex = STATE_WAITING_FOR_CONSUME`.
4. Thread B calls `shutdown()` on the same fd.
5. `Connection::shutdown()` blocks in `queue_and_wait()` before `Operation::Shutdown` can be enqueued.

The exact `SocketWaker::wake()` shutdown early-return is a secondary hazard; in this call path shutdown does not reach `TcpSocket::shutdown()` because the shared reply is already occupied. The current timeout parameter is also ignored, so the wait is effectively indefinite.

## Impact

- Concurrent operations on one fd can deadlock indefinitely.
- A pending accept allocates an `accepted_fd` before waiting, so the fd can remain allocated if the wait is never cancelled.
- Similar behavior can affect blocking `recv()`, `send()`, `recvfrom()`, `sendto()`, `recvmsg()`, and `sendmsg()`.
- Reordering the shutdown waker or resetting the shared futex alone may cause stale operations or deliver the wrong reply to a caller.

## Proposed direction

Refactor the network operation/reply ownership in a follow-up change:

- use a reply channel per operation, or add explicit pending-operation ownership;
- provide cancellation that wakes a blocked operation with `ECONNABORTED`;
- prevent late/stale waker replies from completing a newer operation (generation/token);
- define close/shutdown behavior for pending operations;
- add concurrency tests for all blocking socket operations.

Until that refactor is available, document/enforce one in-flight operation per fd and avoid concurrent use of the same fd from multiple threads.

## References

- PR #454: TCP `accept()` implementation
[...]

PR #461, opened 2026-08-26
net: support socket reuse and I/O timeouts

xuchang-vivo:
## Summary
- Accept and retain `SO_REUSEADDR` as a compatibility option while preserving the current one-listener-per-endpoint behavior.
- Match socket option identifiers exactly and validate option pointers and lengths.
- Give every socket operation its own reply object instead of sharing reply state across calls.
- Enforce `SO_RCVTIMEO` and `SO_SNDTIMEO` for blocking TCP, UDP, and ICMP send/receive operations.
- Treat a zero timeout as clearing the option and reject invalid timeval values.

## Motivation
Rust `TcpListener::bind()` enables `SO_REUSEADDR` before binding. BlueOS previously returned `ENOPROTOOPT`, so otherwise valid listeners failed before reaching `bind()`.

The socket timeout values were stored but never enforced. A blocked operation waited indefinitely, and the connection-wide reply object could retain state across multiple calls. This showed up as agent-loop requests hanging after several HTTP/TLS rounds.

## Reuse-address behavior
smoltcp does not expose endpoint-sharing semantics equivalent to a host TCP stack. This change therefore accepts and reports the option for API compatibility but does not allow multiple listeners to bind the same endpoint. BlueOS still permits one listener per endpoint.

## Timeout and buffer-lifetime safety
Each operation moves through explicit waiting, executing, completed, and cancelled states.

- The network loop must claim a waiting operation before invoking its callback.
- A timeout can cancel only a waiting operation.
- If the network loop is executing a callback, the syscall waits until that execution completes or parks again.
- Blocking socket paths return to the waiting state only after the callback has been transferred to a socket waker.
- Cancelled queued or woken operations are dropped before their callbacks can run.

This prevents a timed-out syscall from returning while the network thread may still access the caller-owned send or receive buffer.

## Related work
- Partially addresses #456 by removing the shared per-connection reply state. Shutdown-driven cancellation of already blocked operations remains follow-up work.
- #459 fixes the target ABI representation used for socket timeval arguments.
- #460 preserves `ConnectionError::Timeout` as `ETIMEDOUT` across the syscall boundary.

This PR does not include BME280 or other sensor changes.

## References
- [POSIX setsockopt()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/setsockopt.html)
- [POSIX getsockopt()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getsockopt.html)

## Testing
- `rustfmt --edition 2021 --check kernel/src/net/connection.rs kernel/src/net/syscalls.rs kernel/src/net/smoltcp/socket/tcp.rs kernel/src/net/smoltcp/socket/udp.rs kernel/src/net/smoltcp/socket/icmp.rs`
- `git diff --check`
- No board build was run.

### DIFF
--- a/kernel/src/net/connection.rs
+++ b/kernel/src/net/connection.rs
@@ -51,9 +51,9 @@ pub struct Connection {
     local_endpoint: Mutex<Option<IpListenEndpoint>>,
     remote_endpoint: Mutex<Option<IpEndpoint>>,
     is_nonblocking: AtomicBool, // default io mode is blocking, use O_NONBLOCK to set non-blocking
+    reuse_address: AtomicBool,
     recv_timeout: Mutex<Option<Duration>>, // block indefinitely as default
     send_timeout: Mutex<Option<Duration>>, // block indefinitely as default
-    ipc_reply: Arc<OperationIPCReply>,
 }
 
 impl Connection {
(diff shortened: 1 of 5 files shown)


xuchang-vivo:
build_prs

PR #462, opened 2026-08-26
 accelerate TCP download

jinwjinl:
Currently, there's a situation where the TCP socket buffer is too small, preventing the application layer's receive window from being set to a larger size. Additionally, under heavy load, the kernel enters a futex state, causing the application layer to be unable to resume downloading.

### DIFF
--- a/kernel/src/net/smoltcp/socket/tcp.rs
+++ b/kernel/src/net/smoltcp/socket/tcp.rs
@@ -69,7 +69,7 @@ impl TcpSocket {
         };
 
         let tcp_socket = {
-            let tcp_rx_buffer = tcp::SocketBuffer::new(vec![0; 1024]);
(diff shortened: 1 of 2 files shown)


jinwjinl:
build_prs

wangchen-vivo:
build_prs

wangchen-vivo:
build_prs

kernel/src/net/smoltcp/socket/tcp.rs
```diff
@@ -69,7 +69,7 @@ impl TcpSocket {
         };
 
         let tcp_socket = {
-            let tcp_rx_buffer = tcp::SocketBuffer::new(vec![0; 1024]);
+            let tcp_rx_buffer = tcp::SocketBuffer::new(vec![0; 8 * 1024]);
```
xuchang-vivo, line 72: Make this can be configured

jinwjinl:
build_prs

jinwjinl:
build_prs

PR #463, opened 2026-08-26
ci: harden cross-repo build workflow against workflow injection

wangchen-vivo:
Fixes the command-injection RCE reported in [vivoblueos/kernel#458](https://github.com/vivoblueos/kernel/pull/458), where attacker-controlled PR content was interpolated into `execSync` shell strings and could execute arbitrary commands in CI.

## Changes

- **Command injection**: run format/license/check commands via `execFileSync` argument arrays instead of `execSync` string concatenation, so filenames/repo names are never interpreted by a shell (`poc$(id).rs` no longer expands).
- **Fail-closed checks**: a format tool that fails to run (file missing, parse error, tool crash) now fails the job instead of being skipped and reported as "All modified files pass format check".
- **Least privilege**: format, license, and build jobs run with `permissions: {}` and a placeholder `github-token: unused`, so injected code cannot obtain the job's `GITHUB_TOKEN`. Container GHCR login credentials and `packages: read` were removed because the image is public.
- **Tokenless sync**: the old `setup` job is split into `resolve_prs`, which keeps a token to comment and resolve PRs, and `apply_prs`, which runs repo init, sync, merge, and upload without one. This way `repo sync` has no `GITHUB_TOKEN`, so a `<repo-hooks>` hook injected via an attacker-controlled manifests PR cannot execute untrusted code.
- **Input validation**: `build_prs` links must belong to this org; PR metadata is passed to shell steps via `env` rather than `${{ }}` interpolation into `run:`, so a value containing shell metacharacters is treated as data and never evaluated as commands.
- **Resource limits**: `concurrency` is grouped per PR, so a new `build_prs` comment cancels the in-progress run instead of launching another set of matrix build jobs.

### DIFF
--- a/.github/workflows/cross_repos_build.yml
+++ b/.github/workflows/cross_repos_build.yml
@@ -4,6 +4,9 @@ on:
   issue_comment:
     types: [created]
 
+permissions:
+  pull-requests: write
+
 jobs:
   call_workflow:
     uses: vivoblueos/kernel/.github/workflows/reusable_cross_repos_build.yml@main
--- a/.github/workflows/cross_repos_build.yml
+++ b/.github/workflows/cross_repos_build.yml
@@ -1,12 +1,9 @@
 name: Trigger build cross multiple repositories
-
 on:
   issue_comment:
     types: [created]
-
 permissions:
   pull-requests: write
-
 jobs:
   call_workflow:
     uses: vivoblueos/kernel/.github/workflows/reusable_cross_repos_build.yml@main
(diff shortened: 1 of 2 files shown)


wangchen-vivo:
build_prs

zhaowenbo-vivo (approved):

wangchen-vivo:
build_prs

wangchen-vivo:
build_prs

wangchen-vivo:
build_prs

han-jiang277 (approved):

PR #464, opened 2026-08-26
net: implement TCP accept and stabilize socket operations

xuchang-vivo:
## Summary
- Implement the TCP `accept()` path and accepted-socket setup.
- Preserve accepted socket ownership and peer endpoint information.
- Retain UDP ephemeral-port leases and use the correct socket type for connect port allocation.
- Add `SO_REUSEADDR` compatibility for the single-listener smoltcp configuration.
- Validate socket option identifiers, timeval layouts, lengths, and values.
- Use per-operation socket reply state and enforce configured blocking I/O timeouts.
- Preserve socket errors, including timeout and would-block conditions, across syscall and VFS boundaries.
- Keep socket shutdown and closed-socket state transitions consistent.

## Relation to issue #456
This partially addresses #456 by removing the shared per-connection reply object and allowing independent socket operations to be queued. Shutdown-driven cancellation of an already blocked operation remains follow-up work.

## Scope
This PR contains the kernel network changes currently on `xc/add_more_caps_to_agent`. It does not include the uncommitted ESP32-C3 board changes, BME280 changes, or application changes.

## Testing
- Ran `rustfmt --edition 2021 --check` on the files changed by the final socket-state/options commit.
- Ran `git diff --check` on the same files.
- No board build was run.

### DIFF
--- a/kernel/src/net/connection.rs
+++ b/kernel/src/net/connection.rs
@@ -344,6 +344,9 @@ impl Connection {
                 let mut endpoint = self.local_endpoint.lock();
                 if endpoint.is_none() {
                     let local_port = PORT_GENERATOR.acquire_port(self.socket_type, 0)?;
+                    self.local_port_lease
+                        .lock()
+                        .replace(PortLease::new(self.socket_type, local_port));
                     endpoint.replace((local_port).into());
                     Some(local_port)
                 } else {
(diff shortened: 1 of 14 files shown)


xuchang-vivo:
build_prs

PR #466, opened 2026-09-02
fix(serial): avoid USB Serial JTAG cold-boot blocking

xuchang-vivo:
## Summary
- Prevent USB-Serial-JTAG console writes from blocking before the host enumerates the device.
- Skip early-console bytes while the USB IN endpoint is busy instead of waiting with interrupts disabled.
- Enable the USB serial console option for ESP32-C3 debug and release defconfigs.

## Scope
- Based directly on `main`.

## Testing
- Not run locally; repository instructions prohibit unrequested build commands.
- Please rely on CI for compilation checks.


xuchang-vivo:
build_prs

han-jiang277 (approved):

PR #472, opened 2026-09-10
kernel: fix ESP32C6 serial driver interrupt starvation on single-core

xuchang-vivo:
Fix ESP32C6 serial driver interrupt starvation on single-core RISC-V platforms.

## Problem

On single-core ESP32C6 platforms, the serial driver was disabling interrupts unconditionally in four functions:
- flush_rx_fifo()
- flush_tx_fifo()  
- trigger_tx_interrupt()
- get_char()

The most critical issue was in trigger_tx_interrupt(), which disables interrupts while draining the entire 128-byte software ring buffer to hardware. With two threads both doing println! (agent + UI), this accumulated interrupt-off time reached hundreds of milliseconds, blocking WiFi beacon interrupts (100ms period) and causing StationBeaconTimeout → TLS handshake failures.

## Solution

Add #[cfg(smp)] guards to irqsave_lock() calls in these four functions. On single-core platforms, Cell/RefCell provides sufficient protection without disabling interrupts. SMP platforms still get full interrupt-disabling protection.

## Testing

Verified with 50-round agent_loop stress tests on ESP32C6 - no more hangs or TLS failures.

## Board Impact

Single-core RISC-V only (ESP32C6, ESP32C3). SMP builds unchanged.


han-jiang277:
build_prs https://github.com/vivoblueos/build/pull/81

PR #474, opened 2026-09-14
Refactor the buddy allocator

YinhuaChen-vivo:
# Refactor the buddy allocator and improve allocation routing

## Summary

This change separates the reusable buddy algorithm from its kernel integration, fixes the routing boundary between the small-object heap and the buddy allocator, simplifies `GlobalAlloc::realloc`, and replaces globally serialized buddy tests with isolated local allocators.

It also adds concurrency coverage for independent buddy allocators and for multiple threads sharing one buddy allocator.

## Motivation

The buddy implementation previously mixed two different responsibilities:

- the generic page allocation algorithm; and
- kernel-specific synchronization, address translation, global state, and boot-time memory layout.

This made the algorithm harder to reuse and forced its tests to operate on the global buddy allocator. Those tests needed an exclusive test lock, so they could not run in parallel.

There was also a routing gap in the global allocator. The small-object allocator supports allocations only up to 2048 bytes, but the old code switched to the buddy allocator only at 4096 bytes. Requests in the range 2049 through 4095 bytes could therefore be sent to an allocator that could not serve them.

## Architecture

The implementation is now split into a reusable algorithm layer and a kernel wrapper layer:

```text
                        GlobalAlloc
                            |
                            v
                 kernel allocation routing
                            |
                            v
        +-------------------------------------------+
        | kernel/src/allocator/buddy                |
        |                                           |
        | BuddyHeap                                 |
        | - SpinLock synchronization                |
        | - PFN <-> physical address translation    |
        | - kernel DRAM and metadata layout         |
        | - global allocator instance               |
        +----------------------+--------------------+
                               |
                               v
        +-------------------------------------------+
        | allocator/src/buddy.rs                    |
        |                                           |
        | BuddyAllocator                            |
        | - free-list management                    |
        | - block splitting and coalescing          |
        | - aligned page allocation                 |
        | - page accounting                         |
        | - caller-provided metadata storage        |
        | - no internal synchronization             |
        +-------------------------------------------+
```

The raw `BuddyAllocator` works with relative page frame numbers and returns an opaque `PageFrame`. The caller owns synchronization, metadata storage, address translation, and metadata lifetime. This keeps kernel policy out of the reusable allocator crate.

The kernel-facing `BuddyHeap` wraps the raw allocator in an IRQ-safe spin lock and stores immutable translation data after initialization. It exposes physical-address and PFN operations while retaining the existing global buddy allocator role.

The old kernel-specific page descriptor module is removed. The allocator-private descriptor now contains only the state required by the buddy algorithm.

## Allocation routing fix

Routing uses the effective request size:

```text
effective_size = max(layout.size(), layout.align())

     effective_size <= 2048       effective_size > 2048
    +------------------------+    +------------------------+
    | small-object heap      |    | buddy allocator        |
    +------------------------+    +------------------------+
```

The boundary is intentionally `> 2048`, not `>= 4096`:

```text
Request size       Old route              New route
------------       ---------              ---------
1..2048            small-object heap      small-object heap
2049..4095         small-object heap      buddy allocator
4096 and above     buddy allocator        buddy allocator
```

Alignment participates in the same decision. A small allocation with page-sized alignment is routed to buddy so that its alignment requirement can be satisfied.

Allocation, deallocation, and reallocation now use the same boundary, preventing a pointer from being returned to a different allocator than the one that created it.

## Reallocation cleanup

`GlobalAlloc::realloc` is reduced to three cases:

```text
old allocation            new allocation             action
--------------            --------------             ------
small-object heap         small-object heap          use heap realloc
buddy, same order         buddy, same order          reuse the block
all other transitions     any different route/order  allocate, copy, free
```

For the general case, `realloc` now delegates to the normal routed `alloc` and `dealloc` implementations:

```text
new_ptr = alloc(new_layout)
if allocation failed:
    keep old_ptr valid and return null
copy min(old_size, new_size) bytes
dealloc(old_ptr, old_layout)
return new_ptr
```

This removes the duplicated small-to-buddy, buddy-to-small, and buddy-to-buddy code paths while preserving the standard failure behavior.

## Parallel and isolated buddy tests

Buddy tests no longer mutate the global buddy allocator directly under an exclusive test lock. Each wrapper test obtains a page-aligned backing arena from the global allocator and initializes a local `BuddyHeap` inside that arena:

```text
global buddy allocator
          |
          +-- alloc arena A
          |        |
          |        +-- local BuddyHeap A
          |             +----------------+----------------------+
          |             | local metadata | local manageable RAM |
          |             +----------------+----------------------+
          |
          +-- alloc arena B
                   |
                   +-- local BuddyHeap B
                        +----------------+----------------------+
                        | local metadata | local manageable RAM |
                        +----------------+----------------------+
```

Because the arenas do not overlap, tests can execute in parallel without sharing allocator state. The fixture also keeps the backing allocation alive until after the local allocator is dropped, which preserves the lifetime of its intrusive free-list metadata.

The raw algorithm tests use their own allocator instances and caller-owned metadata buffers, so they are isolated from both the kernel wrapper and one another.

## Test coverage

The updated test suite covers:

- initialization and invalid metadata inputs;
- arbitrary manageable and reserved page ranges;
- allocation and deallocation at every supported order;
- requested alignment, exhaustion, recovery, splitting, and coalescing;
- page conservation and a deterministic mixed-operation reference model;
- size-to-order calculations at page boundaries;
- PFN and physical-address translation, including bounds and overflow cases;
- real reads and writes through locally backed physical pages;
- routing at 2048 and 2049 bytes, including alignment-based routing;
- reallocation in both directions across the buddy threshold;
- preservation of the old allocation when reallocation fails;
- multiple independent local buddy allocators operating concurrently; and
- multiple threads concurrently allocating from one shared local buddy allocator.

The two concurrency scenarios exercise different guarantees:

```text
Scenario 1: independent state          Scenario 2: shared state

thread A -> local buddy A              thread A --+
                                                   +-> one local BuddyHeap
thread B -> local buddy B              thread B --+    (SpinLock protected)
```

Together, they verify both test isolation and kernel-wrapper synchronization.


YinhuaChen-cloud:
build_prs

YinhuaChen-vivo:
build_prs

YinhuaChen-vivo:
build_prs

YinhuaChen-vivo:
build_prs

YinhuaChen-vivo:
build_prs

kernel/src/allocator/buddy/heap.rs
```diff
@@ -12,533 +12,290 @@
...
-
-#[cfg(test)]
-use crate::scheduler;
-#[cfg(test)]
-use core::sync::atomic::{AtomicUsize, Ordering};
+use core::{cell::UnsafeCell, ptr::NonNull};
 
+/// Virtual address marking the end of the kernel image, provided by the
+/// linker script. It is typically aligned to `PAGE_SIZE`.
 extern "C" {
```
xuchang-vivo, line 23: PLZ change `extern "C"` to `unsafe extern "C"`

YinhuaChen-vivo: Yeah, you're right. Adding unsafe complies with the Rust 2024 Edition rules.

YinhuaChen-vivo: done

YinhuaChen-vivo:
build_prs

PR #477, opened 2026-09-18
boot(aarch64): add arm64 Image header so booti keeps image at linked address

xuchang-vivo:
## Summary
- Emit the 64-byte arm64 Image header (Linux `Documentation/arch/arm64/booting.rst`) at the top of `arch_bootstrap!`: `code0` "MZ", branch over the header, `text_offset`, `image_size`, flags bit3 (keep placement), and the "ARM\x64" magic.
- Add a per-board `TEXT_OFFSET` constant matching each board's linked base (`qemu_virt64_aarch64`: `0x4000_0000 + 0x80000`; `rk3568`: `0xc0_0000 + 0x800`), published in the header so loaders such as U-Boot `booti` keep the image at its linked address instead of relocating it to the base of DRAM.
- Mark `_start` as `-> !` (it never returns), which lets naked-function validation accept the trailing `udf`.

## Scope
- Based directly on `main`.
- Touches aarch64 arch bootstrap + the two aarch64 boards' config; no behavior change for other architectures/boards.

## Testing
- `gn gen` + `ninja default` + `ninja check_all` pass for `qemu_virt64_aarch64` and `rk3568` (debug, direct-syscall).
- Verified the emitted header bytes of both built images with `xxd`: `MZ` at 0x00, `text_offset` (0x80000 / 0x800), `image_size`, flags bit3, `ARM\x64` magic at 0x38.

See https://docs.kernel.org/arch/arm64/booting.html


xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

PR #478, opened 2026-09-18
fix(rk3568): wire console UART interrupt so serial output no longer truncates

xuchang-vivo:
## Problem

On the RK3568 board, serial console output is truncated mid-print. Once kernel output exceeds a single hardware FIFO fill, all further bytes are silently dropped.

## Root Cause

Three defects stacked on the rk3568 board init:

1. **Wrong UART wired to the ISR.** `PL011_UART0_BASE = 0xFDD50000` / `PL011_UART0_IRQNUM = 148` point at **uart0_pmu** (GIC SPI 116 -> INTID 148) — the always-on PMU-domain UART that is *not* the console. The actual serial console is **uart2** @ `0xFE660000` (GIC SPI 118 -> **INTID 150**, level-high), as used by `define_peripheral!(console_uart)` with the hardcoded `0xFE660000` in the same file. The ISR was registered for an interrupt that console traffic can never raise.

2. **GIC never delivers the interrupt.** No `enable_irq_with_priority()` for either the console UART or the generic timer, and no `set_trigger()`. `Serial::send_bytes()` fills the hardware FIFO once via `trigger_tx_interrupt()` and then depends on TX interrupts to drain the 256-byte software ring — an interrupt that is masked at the GIC level never fires, so the ring fills and output stalls.

3. **Timer interrupt equally dead** — same missing-enable problem for INTID 30, which the kernel clock needs.

## Fix

- Rename the misleading `PL011_UART0_*` constants to `CONSOLE_UART_*` (they are DW APB / Ns16x50 registers, not PL011).
- Point them at uart2: base `0xFE660000`, IRQ INTID 150.
- `enable_irq_with_priority()` for both the console UART (150) and generic timer (30) interrupts, `set_trigger(Level)` for the UART (level, active-high per the rk3568 devicetree).
- Register `Ns16x50Isr` against INTID 150.
- Mirror the working pattern from the `qemu_virt64_aarch64` board.
- `define_peripheral!` now uses `config::CONSOLE_UART_BASE` instead of a duplicate hardcoded address.

## Verification

- Kernel crate + `kernel_unittest` + `kernel_integration_test` build clean (full `obj` tree rebuild with rustc 1.84.0-dev).
- Binary-level check on the produced image (`bin/shell.bin` embedded in `image.fit`):
  - `mov w0, #0x96` (150) appears 3× in `rk3568::init`; `bl enable_irq_with_priority` ×2, `bl set_trigger`, `bl register_handler` all present.
  - stale `mov w0, #0x94` (148) appears 0 times.


xuchang-vivo:
build_prs

xuchang-vivo:
build_prs

han-jiang277 (approved):

PR #483, opened 2026-09-21
allocator: reject size overflow in calloc

### LINKED ISSUE #313: allocator.calloc 未对required_size做溢出校验，会导致崩溃
问题：

```
pub fn calloc(count: usize, size: usize) -> *mut u8 {
// 对required_size 未做校验，异常时后面layout代码会崩溃
    let required_size = count * size;
    const ALIGN: usize = core::mem::size_of::<usize>();
    let layout = Layout::from_size_align(required_size, ALIGN).unwrap();
    if let Some(alloc_ptr) = HEAP.alloc(layout) {
        unsafe { ptr::write_bytes(alloc_ptr.as_ptr(), 0, required_size) };
        alloc_ptr.as_ptr()
    } else {
        ptr::null_mut()
    }
}
```

虽然calloc未暴露给应用层，但是其风险存在，可通过增加相关系统调用calloc暴露给应用层进行演示：
```
> calloc 1024 8
Allocating zero-initialized memory for 1024 elements of 8 bytes each...
Total size: 8192 bytes
Allocation successful!
  Address: 0x2102e428
  Count: 1024
  Size per element: 8
  Total bytes: 8192
  [VERIFIED] First 64 bytes are zero-initialized
> calloc 1024 800000000
Allocating zero-initialized memory for 1024 elements of 800000000 bytes each...
Total size: 3156213760 bytes
thread 'shell' panicked at ../../kernel/kernel/src/allocator/mod.rs:135:64:
called `Result::unwrap()` on an `Err` value: LayoutError
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'shell' panicked at ../../kernel/kernel/src/arch/arm/hardfault.rs:220:5:

        ==== HARD FAULT ====
        FRAME: IsrContext {r0: 0x2102b9c0 r1: 0x95e1c r2: 0x3 r3: 0x2 r12: 0x21017200 lr: 0x6bf0d pc: 0x465ac xpsr: 0x21000000 }
        FAULT REGS: 
HFSR: 0x40000000
  - Forced Hard Fault
CFSR: 0x00010000
Fault Status:
  Usage Fault:
    - Undefined instruction
AFSR: 0x00000000

        XPSR: XPSR: 0x20000003
  Exception Number: 3
  - Invalid state
  - Branch target identification active
  IT/ICI/ECI flag: false
  IT/ICI/ECI value: 0
  Condition flags:
    N=0 Z=0 C=1 V=0 Q=0

        
stack backtrace:
panicked at ../../librs/src/lib.rs:98:5:
not yet implemented
thread panicked while processing panic. aborting.
qemu: fatal: Lockup: can't escalate 3 to HardFault (current priority -1)

R00=00000004 R01=00000011 R02=00000002 R03=00000001
R04=21fffc68 R05=00095e00 R06=00000001 R07=21fffbe8
R08=00000001 R09=21fffe1c R10=21fffdfc R11=21fffd70
R12=21017200 R13=21fffbe8 R14=00066e0d R15=000465ac
XPSR=21000003 --C- T handler
FPSCR: 00000000
已中止 (核心已转储)
```

ConradLu2740:
allocator: reject size overflow in calloc

## Problem

`calloc` computed `required_size = count * size` with a wrapping
multiplication. On 32-bit targets, products such as
`1024 * 800000000` or `3 * 0x60000000` wrap around to a smaller value.
Depending on the wrapped result this either panicked while building the
`Layout` (issue #313) or, worse, produced a valid but undersized layout,
silently under-allocating the block the caller believes it owns - a heap
overflow waiting to happen.

## Fix

Use `checked_mul` and return a null pointer on overflow, matching the C
standard's requirement that `calloc` detect multiplication overflow and
report failure.

## Testing

- Added regression test `calloc_rejects_size_overflow` in
  `kernel/src/allocator/mod.rs`, covering the plain overflow case, the
  wrapped-but-valid-layout case, and a zero-initialization sanity check.
- Verified the test fails on the pre-fix code (assertion panics on the
  `calloc(2, usize::MAX / 2 + 1)` case) and passes with the fix.
- `ninja -C out/shell_test kernel/kernel:run_unittest`: all 184
  unittests pass on qemu_mps2_an385 (thumbv7m, 32-bit).

Fixes #313


ConradLu2740:
build_prs https://github.com/vivoblueos/kernel/pull/483

